PR #34, opened 2025-09-05
feat(cli): add threaded Ctrl-P pause listener for synchronous SDK

malhotra5:
This PR adds a background key listener to support pausing the agent in the synchronous SDK model.

Key changes
- Introduce PauseListener (threading.Thread) using prompt_toolkit raw input
  - Listens for Ctrl-P and triggers a pause callback
  - Prints a short notification and uses a print lock to avoid output interleaving
- ConversationRunner lifecycle management
  - Starts the listener during run() cycles so Ctrl-P can pause while the agent is executing
  - Stops the listener prior to interactive confirmation prompts
  - Ensures the listener is stopped on exit
- Pause callback attempts conversation.pause() first, falling back to agent.pause() if needed
- Retains fully synchronous CLI flow; no asyncio or pub/sub needed

Notes
- Only Ctrl-P is intercepted to avoid conflicting with Ctrl-C (KeyboardInterrupt).
- The implementation assumes conversation.pause()/agent.pause() are thread-safe; otherwise we can swap to a signaling flag read between run() calls.

Testing
- `pytest` passes locally (24 tests).

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/d8ae764387c64f558e2b829b8e02ff48)


malhotra5:
@openhands read the diff of this pr

after the conversation has been paused make it so that /resume will restart the conversation.run loop

malhotra5:
@openhands currently we only resume if `/resume` command is typed; however if the user types "please continue" or any follow up message for that matter start the run till completion loop again

malhotra5:
@openhands the initial specs have changes which has broken our unit tests

please fix the failing unit tests and only modify the test files, nothing else

make sure you remove `openhands/code` from your python before you run the unit tests

malhotra5:
@openhands fix the failing unit tests; only modify the unit test file and nothing else, the initial specs have changed

malhotra5:
@openhands please create a test file for the pause functionality; write one test for the following cases 

1. if you pause in confirmation mode then you aren't shown the confirmation selection, you simply return to the main input page

2. if you pause when not in confirmation mode, your agent step wraps up and returns to the main input page

3. if you keyboard interrupt during confirmation mode, it will pause the agent and return to the main input page

4. if you try resume or any message from the main input page, it will resume the agent 

5. if you try resume or any message from the main input page with confirmation mode, and the user had paused before seeing the confirmation selection for the action, it will resume and show the confirmation options

malhotra5:
@openhands look at `tests/test_pause_functionality.py` 

I've fixed `TestPauseListener` for you; you are able to pass keyboard inputs to simulate pausing

please rework the other tests, they're not actually testing pausing. make sure that the conversation object is real and its running (mock the llm responses); make sure that the you simulate the pause input and see that the conversation runner will pause the conversation object properly

PR #56, opened 2025-11-13
Only set litellm_extra_body for litellm_proxy models

xingyaoww:
## Problem
Some LLM providers like Cerebras don't support the `extra_body` parameter. Currently, we always set `litellm_extra_body` for all models, which causes issues when using providers that don't support it.

## Solution
This PR ensures that `litellm_extra_body` is only set when the model name contains `"litellm_proxy"`, preventing issues with providers that don't support extra_body parameters.

## Changes
- Added `should_set_litellm_extra_body()` helper function to check if model name contains `"litellm_proxy"`
- Updated all LLM instantiation points to conditionally set `litellm_extra_body`:
  - `build.py`: dummy model for testing
  - `settings_screen.py`: when saving LLM settings  
  - `store.py`: when loading agent and condenser LLMs
- Added tests for the new helper function

## Testing
- All existing tests pass
- Added new unit tests for `should_set_litellm_extra_body()` function

## Related
This PR should be merged **before** https://github.com/OpenHands/software-agent-sdk/pull/1153 which makes the corresponding changes in the SDK to always forward `litellm_extra_body` when provided (removing the proxy-only restriction there).

cc @neubig @simonrosenberg @li-boxuan

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/a97333757286405093f10fadb00caf72)

### DIFF
--- a/openhands_cli/tui/settings/settings_screen.py
+++ b/openhands_cli/tui/settings/settings_screen.py
@@ -186,9 +186,9 @@ class SettingsScreen:
         )
 
     def _save_llm_settings(self, model, api_key, base_url: str | None = None) -> None:
-        kwargs: dict[str, Any] = {}
+        extra_kwargs: dict[str, Any] = {}
         if should_set_litellm_extra_body(model):
-            kwargs["litellm_extra_body"] = {
+            extra_kwargs["litellm_extra_body"] = {
                 "metadata": get_llm_metadata(model_name=model, llm_type="agent")
             }
         llm = LLM(
@@ -196,7 +196,7 @@ class SettingsScreen:
             api_key=api_key,
             base_url=base_url,
             usage_id="agent",
-            **kwargs,
+            **extra_kwargs,
         )
 
         agent = self.agent_store.load()
--- a/build.py
+++ b/build.py
@@ -277,14 +277,14 @@ def main() -> int:
     # Test the executable
     if not args.no_test:
         model_name = "dummy-model"
-        kwargs: dict[str, Any] = {}
+        extra_kwargs: dict[str, Any] = {}
         if should_set_litellm_extra_body(model_name):
-            kwargs["litellm_extra_body"] = {
+            extra_kwargs["litellm_extra_body"] = {
                 "metadata": get_llm_metadata(
                     model_name=model_name, llm_type="openhands"
                 )
             }
-        llm = LLM(model=model_name, api_key="dummy-key", **kwargs)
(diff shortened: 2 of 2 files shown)


xingyaoww:
This PR addresses the CLI side of the issue mentioned in https://github.com/OpenHands/software-agent-sdk/pull/1153.

The strategy is:
1. **SDK side** (PR #1153): Always forward `litellm_extra_body` when provided, letting downstream providers handle/validate it
2. **CLI side** (this PR): Only set `litellm_extra_body` for `litellm_proxy` models

This ensures:
- Providers like Cerebras that don't support `extra_body` won't receive it
- `litellm_proxy` models continue to get the metadata they need
- The SDK remains flexible and doesn't impose CLI-specific restrictions

openhands_cli/tui/settings/settings_screen.py
```diff
@@ -181,15 +185,23 @@ def handle_advanced_settings(self, escapable=True):
...
+                    "metadata": get_llm_metadata(model_name=model, llm_type="agent")
+                },
+            )
+        else:
+            llm = LLM(
+                model=model,
+                api_key=api_key,
+                base_url=base_url,
+                usage_id="agent",
+            )
```
xingyaoww: @OpenHands simplify the code here by putting litellm_extra_body to kwargs and pass **kwargs to LLM constructor

openhands_cli/utils.py
```diff
@@ -7,6 +7,22 @@
...
+    Only set litellm_extra_body for litellm_proxy models to avoid issues
+    with providers that don't support extra_body parameters.
+
+    Args:
+        model_name: Name of the LLM model
+
+    Returns:
+        True if litellm_extra_body should be set, False otherwise
+    """
+    return "litellm_proxy" in model_name
```
enyst, line 23: Shouldn’t this be `openhands/ in model_name` actually? 🤔 
I _think_ in CLI, the user chose `openhands/`, and in the SDK, we translate that to a `litellm_proxy/`, unless I’m missing something?
https://github.com/OpenHands/software-agent-sdk/blob/57b365759656cf382a2731fd04fa550e45eabc63/openhands-sdk/openhands/sdk/llm/llm.py#L339

xingyaoww: @enyst dang you are absolutely right 😓 

@OpenHands pls create a new PR that check if we need to set extra body by checking for "openhands/" in model name

PR #64, opened 2025-11-16
Update CLI visualizer to use horizontal rules instead of panels

xingyaoww:
## Summary

This PR updates the CLI visualizer to use the improved visualization approach from the agent-sdk's default visualizer, making it easier to copy/paste agent output without dealing with panel border characters.

image

## Changes

- **Replace Panel-based rendering with Rule (horizontal rule) headers**
  - Switched from `Panel` to `Rule` for event headers
  - Content is now displayed without surrounding borders
  - Makes copy/paste much easier (addresses the complaint mentioned in [agent-sdk#1170](https://github.com/OpenHands/software-agent-sdk/pull/1170))

- **Introduce EventVisualizationConfig for cleaner event handling**
  - Uses a declarative configuration approach for event types
  - Reduces the large `isinstance` chain to a clean lookup table
  - Makes it easier to add new event types

- **Use build_event_block() helper for consistent rendering**
  - Centralizes event block construction
  - Ensures consistent spacing and formatting

- **Preserve CLI-specific features**
  - Keeps the `name` parameter for multi-agent scenarios
  - Maintains all existing functionality (highlighting, metrics, etc.)

- **Updated tests**
  - Modified tests to work with `Group` instead of `Panel`
  - Tests now check for `Rule` elements in the rendered output

## Motivation

As noted in [agent-sdk#1170 review comment](https://github.com/OpenHands/software-agent-sdk/pull/1170#pullrequestreview-3467626364), the CLI uses a custom visualizer. This PR brings the improvements from the SDK's default visualizer to the CLI, particularly the easier copy/paste experience.

## Related

- OpenHands/software-agent-sdk#1170

### Before (Panels)
Content is surrounded by borders, making copy/paste include unwanted characters.

### After (Rules)
Content uses horizontal rule headers, making it easy to copy just the text.

---

**Co-authored-by:** openhands <openhands@all-hands.dev>

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/e967e810865e4eab92be97be57f78d5d)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,8 +18,8 @@ classifiers = [
 # Using Git URLs for dependencies so installs from PyPI pull from GitHub
 # TODO: pin package versions once agent-sdk has published PyPI packages
 dependencies = [
-  "openhands-sdk==1.1",
-  "openhands-tools==1.1",
+  "openhands-sdk==1.2",
+  "openhands-tools==1.2",
   "prompt-toolkit>=3",
   "typer>=0.17.4",
 ]
@@ -126,6 +126,6 @@ typeCheckingMode = "standard"
 
 # USE EXACT COMMIT FROM AGENT-SDK (PR #1170 - visualization improvements)
 
-[tool.uv.sources]
-openhands-sdk = { git = "https://github.com/OpenHands/software-agent-sdk.git", subdirectory = "openhands-sdk", rev = "22773b4a1e5b28c99f5b7bac082195745a307441" }
-openhands-tools = { git = "https://github.com/OpenHands/software-agent-sdk.git", subdirectory = "openhands-tools", rev = "22773b4a1e5b28c99f5b7bac082195745a307441" }
+# [tool.uv.sources]
+# openhands-sdk = { git = "https://github.com/OpenHands/software-agent-sdk.git", subdirectory = "openhands-sdk", rev = "22773b4a1e5b28c99f5b7bac082195745a307441" }
+# openhands-tools = { git = "https://github.com/OpenHands/software-agent-sdk.git", subdirectory = "openhands-tools", rev = "22773b4a1e5b28c99f5b7bac082195745a307441" }
--- a/uv.lock
+++ b/uv.lock
@@ -558,6 +558,18 @@ wheels = [
 ]
 
 [[package]]
+name = "deprecation"
+version = "2.1.0"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "packaging" },
+]
+sdist = { url = "https://files.pythonhosted.org/packages/5a/d3/8ae2869247df154b64c1884d7346d412fed0c49df84db635aab2d1c40e62/deprecation-2.1.0.tar.gz", hash = "sha256:72b3bde64e5d778694b0cf68178aed03d15e15477116add3fb773e581f9518ff", size = 173788, upload-time = "2020-04-20T14:23:38.738Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/02/c3/253a89ee03fc9b9682f1541728eb66db7db22148cd94f89ab22528cd1e1b/deprecation-2.1.0-py2.py3-none-any.whl", hash = "sha256:a10811591210e1fb0e768a8c25517cabeabcba6f0bf96564f8ff45189f90b14a", size = 11178, upload-time = "2020-04-20T14:23:36.581Z" },
+]
+
+[[package]]
 name = "distlib"
 version = "0.4.0"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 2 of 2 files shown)


xingyaoww:
@OpenHands please fix the failing actions on PR #64 at branch `update-visualizer-to-use-rules`

simonrosenberg:
@xingyaoww can we keep the symmetry:
------------ Agent Name ---------------
and not 
Agent Name --------------------------

I think I saw the latter somewhere

openhands_cli/tui/visualizer.py
xingyaoww, line 1: @OpenHands can you simplify the implementation here? why do we need a custom visualizer here? I think we can import stuff like EventVisualizationConfig and indent_content directly from the SDK. 

You can check this PR and see what is available there https://github.com/OpenHands/software-agent-sdk/pull/1170

pyproject.toml
```diff
@@ -124,8 +124,8 @@ pythonVersion = "3.12"
 useLibraryCodeForTypes = true
 typeCheckingMode = "standard"
 
-# UNCOMMENT TO USE EXACT COMMIT FROM AGENT-SDK
+# USE EXACT COMMIT FROM AGENT-SDK (PR #1170 - visualization improvements)
 
-# [tool.uv.sources]
-# openhands-sdk = { git = "https://github.com/OpenHands/agent-sdk.git", subdirectory = "openhands-sdk", rev = "7b695dc519084e75c482b34473e714845d6cef92" }
-# openhands-tools = { git = "https://github.com/OpenHands/agent-sdk.git", subdirectory = "openhands-tools", rev = "7b695dc519084e75c482b34473e714845d6cef92" }
+[tool.uv.sources]
```
malhotra5: we should avoid merging any prs with direct dependencies

it halts the release process

openhands_cli/tui/visualizer.py
```diff
@@ -1,312 +1,22 @@
...
-        parts: list[str] = []
-        parts.append(f"[cyan]↑ input {input_tokens}[/cyan]")
-        parts.append(f"[magenta]cache hit {cache_rate}[/magenta]")
-        if reasoning_tokens > 0:
-            parts.append(f"[yellow] reasoning {abbr(reasoning_tokens)}[/yellow]")
-        parts.append(f"[blue]↓ output {output_tokens}[/blue]")
-        parts.append(f"[green]$ {cost_str}[/green]")
+# CLI-specific customization: skip SystemPromptEvent
+# (not needed in CLI output, only relevant for SDK internals)
+EVENT_VISUALIZATION_CONFIG[SystemPromptEvent] = EventVisualizationConfig(
```
malhotra5, line 18: this lgtm! out if curiosity how likely are changes upstream likely to affect this cli visualizer?

we have new cli designs coming in as well, which are more end user focused and different from agent-sdk visualizer

should we maintain a separate cli visualizer (even if it dupes code in the short term)?

pyproject.toml
xingyaoww: yep, I'm testing the critic, once that's merged, i will look into a SDK release and update this PR

openhands_cli/tui/visualizer.py
xingyaoww: I guess we can consider rebuild the visualizer when we actually need to implement those designs? 👀 

Currently we don't really need to add anything custom to get the visualizer working

malhotra5: Gotcha thanks - just so we're synced on next steps we have the following CLI UI changes planned, lmk if you'd like to take these changes into agent-sdk as well or whether they should remain CLI specific

1. Shows a ticking clock for commands, so that there is a visualizer indicator for longer running commands

2. remove token cost information on each message by default - have a toggle to display it 

3. reduce verbosity of message - collapse command output and `ctrl + t` to expand

enyst: I think point 2 is SDK as well. (and they all sound good 😅 )

I think it makes sense to keep/make a custom CLI visualizer, even if there's some redundancy. It's easier to go back and refactor some into the SDK from CLI, if we want it there. 🤔 

(The reverse might be more time-consuming, it feels like it might imply waiting for SDK releases in order to make / fix stuff in CLI ? Or even being permanently aware of what the default visualizer does, outside its interface...)

xingyaoww: I think we can merge this PR as is, and we can add a custom visualizer back when we actually need it.

Right now, we don't actually need a custom visualizer, so it is probably best to delete it and add it back when needed

IMO adding is easier than modifying the existing structure (especially if it is very redundant and can get out of sync from upstream)

xingyaoww: i'm not really against using a custom visualizer for CLI -- I just think it is probably not necessary to put it in now.

If we are making bigger UI changes, we are likely to completely rewrite the visualizer (i.e., not inheriting from SDK at all!), so deleting code feels easier for now

enyst: Fair enough, we should see what a custom visualizer can reuse and what it's unlikely to, in other words what interface does the default visualizer actually provide to be extended, rather than replaced... if that will be possible.

Issue #66, opened 2025-09-29
[Feature Request] ACP support


kolya-ay:
**What problem or use case are you trying to solve?**

I want to launch OpenHands coding agent in Zed editor through Agent Client Protocol, see:

https://zed.dev/blog/claude-code-via-acp
https://agentclientprotocol.com/

**Additional context**

Vim and other non-VSCode editor's users [can benefit](https://github.com/zed-industries/agent-client-protocol?tab=readme-ov-file#editors) from ACP support too

jpelletier1:
We are considering this - not yet on the roadmap but looking into it!

neubig:
Agreed, this is interesting!

raymyers:
I'm a fan of Zed as well :)

OpenHands "[Agent SDK](https://github.com/All-Hands-AI/agent-sdk)" is being extracted into a separate codebase, which I think will make it easier to implement things like this. If someone were interested in trying I could answer questions as they go.

Some notes:

ACP has recently been brought into the A2A project under the Linux Foundation, but it looks like the A2S Python SDK doesn't support it. https://github.com/a2aproject/a2a-python

This is BeeAI's ACP to A2A migration path which seems to depend on their framework (they previous had the ACP Python SDK). I'm not sure if we want to depend on BeeAI in a main project but if someone wanted to get started then trying this with Agent-SDK in a usage example could be interesting.
https://github.com/i-am-bee/beeai-platform/blob/main/docs/community-and-support/acp-a2a-migration-guide.mdx

xingyaoww:
Let's move this issue to agent-SDK for implementation!

Take a quick look and it seems we already have these interface implemented in https://github.com/All-Hands-AI/agent-sdk/tree/main/openhands/agent_server

xingyaoww:
@OpenHands please implement also the baseline methods for AgentClientProtocol - check documentations here: https://agentclientprotocol.com/protocol/overview.md

```
### Baseline Methods

<ResponseField name="initialize" post={[<a href="./schema#initialize">Schema</a>]}>
  [Negotiate versions and exchange capabilities.](./initialization).
</ResponseField>

<ResponseField name="authenticate" post={[<a href="./schema#authenticate">Schema</a>]}>
  Authenticate with the Agent (if required).
</ResponseField>

<ResponseField name="session/new" post={[<a href="./schema#session%2Fnew">Schema</a>]}>
  [Create a new conversation session](./session-setup#creating-a-session).
[...]

PR #69, opened 2025-11-17
Initial ACP Implementation

xingyaoww:
Overview of ACP: https://agentclientprotocol.com/protocol/overview

**Done in this PR**
- [X] Initialization: https://agentclientprotocol.com/protocol/initialization
- [X] Session Setup: https://agentclientprotocol.com/protocol/session-setup
- [X] Content (including embededContent): https://agentclientprotocol.com/protocol/content#embedded-resource
- [X] Basic Prompt turns (except permission) https://agentclientprotocol.com/protocol/prompt-turn
- [X] Basic Tool Calls: https://agentclientprotocol.com/protocol/tool-calls
- [X] Agent Plan https://agentclientprotocol.com/protocol/agent-plan

**Next**
- Confirmation: https://agentclientprotocol.com/protocol/tool-calls, https://agentclientprotocol.com/protocol/prompt-turn
- Slash Command (for changing confirmation mode): https://agentclientprotocol.com/protocol/slash-commands

**Not planned**
- Using client terminal: https://agentclientprotocol.com/protocol/terminals
- Using client file system: https://agentclientprotocol.com/protocol/file-system
- Using session modes: https://agentclientprotocol.com/protocol/session-modes

Demo Video!

https://github.com/user-attachments/assets/5b921c1d-7543-4d59-b7dd-a6cb51321fd5

Toad is also working:

image

Fixes: https://github.com/OpenHands/OpenHands-CLI/issues/66

### DIFF
--- a/openhands_cli/acp_impl/event.py
+++ b/openhands_cli/acp_impl/event.py
@@ -52,7 +52,7 @@ from openhands.tools.task_tracker.definition import (
     TaskTrackerObservation,
     TaskTrackerStatusType,
 )
-from openhands.tools.terminal.definition import ExecuteBashAction
+from openhands.tools.terminal.definition import TerminalAction
 
 
 logger = get_logger(__name__)
@@ -223,7 +223,7 @@ class EventSubscriber:
                     else:
                         tool_kind = "edit"
                         title = f"Editing {event.action.path}"
-                elif isinstance(event.action, ExecuteBashAction):
+                elif isinstance(event.action, TerminalAction):
                     title = f"{event.action.command}"
                 elif isinstance(event.action, TaskTrackerAction):
                     title = "Plan updated"


xingyaoww:
@OpenHands please fix the failing actions on PR #69 at branch `xw/acp-simplification`

openhands_cli/simple_main.py
```diff
@@ -36,6 +36,12 @@ def main() -> None:
             from openhands_cli.gui_launcher import launch_gui_server
 
             launch_gui_server(mount_cwd=args.mount_cwd, gpu=args.gpu)
+        elif args.command == "acp":
+            import asyncio
+
+            from openhands_cli.acp_impl.main import run_acp_agent
+
+            asyncio.run(run_acp_agent())
```
malhotra5: We were trying to have the CLI only use threading (for various reasons) - I suppose its hard to avoid given the acp library implementation?

openhands_cli/acp_impl/agent.py
```diff
@@ -0,0 +1,600 @@
...
+
+        return conversation
+
+    async def initialize(self, params: InitializeRequest) -> InitializeResponse:
+        """Initialize the ACP protocol."""
+        logger.info(f"Initializing ACP with protocol version: {params.protocolVersion}")
+
+        # Check if agent is configured
+        try:
+            load_agent_specs()
```
malhotra5, line 202: So the agent needs to be setup first via the openhands cli before ACP handshake can be made? 

Is it possible to configure the settings for openhands cli from within the client (e.g zed UI)?

openhands_cli/acp_impl/agent.py
```diff
@@ -0,0 +1,600 @@
...
+        )
+
+    async def authenticate(
+        self, params: AuthenticateRequest
+    ) -> AuthenticateResponse | None:
+        """Authenticate the client (no-op for now)."""
+        logger.info(f"Authentication requested with method: {params.methodId}")
+        return AuthenticateResponse()
+
+    async def newSession(self, params: NewSessionRequest) -> NewSessionResponse:
```
malhotra5, line 236: Does ACP client have an error response object? If this fails will it happen silently with no error explanation?

openhands_cli/acp_impl/agent.py
```diff
@@ -0,0 +1,600 @@
...
+            # Convert ACP prompt format to OpenHands message content
+            message_content = convert_acp_prompt_to_message_content(params.prompt)
+
+            if not message_content:
+                return PromptResponse(stopReason="end_turn")
+
+            # Send the message with potentially multiple content types
+            # (text + images)
+            message = Message(role="user", content=message_content)
+            conversation.send_message(message)
```
malhotra5, line 301: Do we want to run this in a thread? Does ACP support multiple concurrent agent chats?

openhands_cli/acp_impl/agent.py
```diff
@@ -0,0 +1,600 @@
...
+                        content=TextContentBlock(type="text", text=f"Error: {str(e)}"),
+                    ),
+                )
+            )
+            # Check if error was due to cancellation
+            if session_id in self._cancelled_sessions:
+                return PromptResponse(stopReason="cancelled")
+            return PromptResponse(stopReason="end_turn")
+
+    async def _run_with_confirmations(
```
malhotra5: This seems to be a dup of the methods available in the `ConversationRunner` 

I wonder if we can reuse the logic here? Possibly configure it to run in "ACP" mode which returns the required stop reasons

Alternatively maybe we extend the existing class and minimally override with ACP logic

xingyaoww: OH actually i forgot to delete these code. I initially try to implement confirmation in this PR then realize there's too much code duplication. Hence i decided to reduce the scope of this PR - we should really do a refactor of these confirmation logic first before try to implment this w/ ACP

openhands_cli/simple_main.py
xingyaoww: yes, and i don't think we need to worry about it, since CLI is either in "tui" mode OR "acp" mode, it is not possible for both mode to activate.

openhands_cli/acp_impl/agent.py
xingyaoww: > Is it possible to configure the settings for openhands cli from within the client (e.g zed UI)? 

This is not officially supported w/ ACP - and how you configure the setting will likely changes as you switch to a different ACP client

openhands_cli/acp_impl/agent.py
xingyaoww: Doesn't seems so
https://agentclientprotocol.com/protocol/session-setup

openhands_cli/acp_impl/agent.py
xingyaoww: deleted

openhands_cli/acp_impl/agent.py
xingyaoww: Yeah it is possible - and ACP client uses asyncio across the board so it'll probably be good to follow suit

openhands_cli/acp_impl/agent.py
xingyaoww: Actually it seems it does has JSON RPC error object: https://agentclientprotocol.com/protocol/schema#error

Let me see how to get that working

malhotra5: The SDK also has known error types ` https://docs.openhands.dev/sdk/guides/llm-error-handling#quick-start:-using-agents-and-conversations`

It would be great to handle these and convert to ACP error schema

openhands_cli/acp_impl/agent.py
malhotra5: Noting that we should switch to textual library for CLI UI, which also uses asyncio - so it seems we'll use asyncio exclusively in the future and potentially can share some core logic for running the agent between the regular CLI and ACP variant

openhands_cli/acp_impl/agent.py
```diff
@@ -0,0 +1,565 @@
...
+
+    async def cancel(self, params: CancelNotification) -> None:
+        """Cancel the current operation."""
+        logger.info(f"Cancel requested for session: {params.sessionId}")
+
+        try:
+            # Get active conversation and pause it
+            conversation = self._get_or_create_conversation(session_id=params.sessionId)
+            # Pause the conversation (state is preserved in cache)
+            conversation.pause()
```
malhotra5, line 367: The pause event is sent immediately but the agent can take a while to stop

Maybe we want to make sure that async thread for the conversation has terminated before sending a response, so that we're being faithful about the actual conversation status

openhands_cli/acp_impl/utils/convert.py
```diff
@@ -0,0 +1,52 @@
...
+    ImageContentBlock as ACPImageContentBlock,
+    ResourceContentBlock as ACPResourceContentBlock,
+    TextContentBlock as ACPTextContentBlock,
+)
+
+from openhands.sdk import ImageContent, TextContent
+from openhands_cli.acp_impl.utils.resources import convert_resources_to_content
+
+
+def convert_acp_prompt_to_message_content(
```
malhotra5, line 68: nit: we seem to have two conventions `convert_{xyz}` and `transform_{xyz}` for util function names

It would be great if we're consist with them, and organize them appropriately as well. it seems this file is specifically for messages related conversions

openhands_cli/acp_impl/agent.py
```diff
@@ -0,0 +1,604 @@
...
+        # Get the current event loop for the callback
+        loop = asyncio.get_event_loop()
+
+        def sync_callback(event: Event) -> None:
+            """Synchronous wrapper that schedules async event handling."""
+            asyncio.run_coroutine_threadsafe(subscriber(event), loop)
+
+        # Create conversation with persistence support
+        # The SDK automatically loads from disk if conversation_id exists
+        conversation = Conversation(
```
malhotra5, line 180: Should  we disable the default visualizer ACP conversations?

openhands_cli/acp_impl/agent.py
```diff
@@ -0,0 +1,604 @@
...
+        # Check if agent is configured
+        try:
+            load_agent_specs()
+            auth_methods = []
+            logger.info("Agent configured, no authentication required")
+        except MissingAgentSpec:
+            # Agent not configured - this shouldn't happen in production
+            # but we'll return empty auth methods for now
+            auth_methods = []
+            logger.warning("Agent not configured - users should run 'openhands' first")
```
malhotra5, line 209: TODO: send error response here via ACP protocol for failing to initialize agent

malhotra5: What happens if the user has configured oauth based MCPs? What does the user journey look like?

raymyers:
> Coverage 82%

You love to see it :)

neubig:
Just FYI, copy-pasting a screenshot in to the ACP seems to cause it to crash. But I think we could open an issue and fix it separately after this is merged.

Screenshot 2025-11-22 at 7 34 38 AM

neubig:
I'm also getting this error:

```
Internal error: {
  "reason": "Failed to process prompt",
  "details": "Conversation run failed for id=88f7a3d7-49d7-471d-874c-d9e078ae7a4a: [Errno 35] write could not complete without blocking"
}
```

neubig:
I've investigated the `[Errno 35] write could not complete without blocking` error and created a detailed issue to track this: #130

## Summary

This is a stdio buffer overflow issue that occurs when the agent generates events faster than the ACP client can read them from stdout. The stdout buffer fills up, and writes fail with `EAGAIN`/`EWOULDBLOCK` (errno 35 on macOS/BSD).

This is a well-known issue with stdio-based protocols - similar problems have been reported in:
- MCP Python SDK: [modelcontextprotocol/python-sdk#1333](https://github.com/modelcontextprotocol/python-sdk/issues/1333)
- Other projects using stdio for IPC

## Root Cause

The `EventSubscriber` in `event.py` converts every SDK event into ACP `sessionUpdate()` calls that write to non-blocking stdout. During high activity (many tool calls, large outputs), the rate of writes exceeds the client's read rate, causing buffer overflow.

## Recommended Solution

See issue #130 for detailed solutions. The immediate fix would be:
1. Add error handling to catch `BlockingIOError` in `EventSubscriber`
2. Implement retry logic with backpressure
3. Add event queuing/throttling to prevent rapid-fire updates

Please see #130 for full technical details, architecture analysis, and implementation recommendations.

xingyaoww:
@OpenHands merge from main and resolve all the conflicts. Also, when we launch ACP, can you try to redirect all stdout produced from the program (not the `reader, writer = await stdio_streams()`, but the other prints statement that might make it to the stdout/stderr) to a log file in CWD (`openhands_acp_YYYYMMDD-HHMM.log`) when `DEBUG` is true (from openhands.sdk.logger import DEBUG), and into dev null when DEBUG is false

openhands_cli/acp_impl/agent.py
xingyaoww: Yep, already handled all exception and convert them to RequestError

openhands_cli/acp_impl/agent.py
xingyaoww: @OpenHands please address this

openhands_cli/acp_impl/utils/convert.py
xingyaoww: @OpenHands please address this comment

openhands_cli/acp_impl/utils/resources.py
```diff
@@ -0,0 +1,101 @@
...
+)
+
+from openhands.sdk import ImageContent, TextContent
+from openhands.sdk.context import Skill
+
+
+RESOURCE_SKILL = Skill(
+    name="user_provided_resources",
+    content=(
+        "You may encounter sections labeled as user-provided additional context or resources. These blocks contain files or data that the user referenced in their message. They may include plain text, images, code snippets, or binary content saved to a temporary file. Treat these blocks as part of the user’s input. Read them carefully and use their contents when forming your reasoning or answering the query. If a block points to a saved file, assume it contains relevant binary data that could not be displayed directly."  # noqa: E501,
```
xingyaoww: > when there is a None trigger does this get extended into the system prompt?

Yes, that's the default behavior here!

openhands_cli/acp_impl/agent.py
xingyaoww: I think the logs will just be printed to the stdout, and user can't see them, then the ACP will timeout

openhands_cli/acp_impl/agent.py
xingyaoww: done

xingyaoww:
@OpenHands please fix the failing actions on PR #69 at branch `xw/acp-simplification`

openhands_cli/acp_impl/utils/resources.py
```diff
@@ -0,0 +1,110 @@
...
+            )
+        )
+
+    elif isinstance(res, ACPBlobResourceContents):
+        mime_type = res.mimeType or ""
+
+        # 1. If it's an image, directly return ImageContent
+        if mime_type.startswith("image/"):
+            data_uri = f"data:{mime_type};base64,{res.blob}"
+            return ImageContent(image_urls=[data_uri])
```
xingyaoww: @OpenHands please help me write a converter function that converts "mime_type" + "blob" into a format that supported by LLM API. I've got the following error:

```
Error: Conversation run failed for id=800b83c1-d74a-4a48-99d3-db0b47a6befc: litellm.BadRequestError: Error code: 400 - {'error': {'message': 'litellm.BadRequestError: AnthropicException - {"type":"error","error":{"type":"invalid_request_error","message":"messages.0.content.0.image.source.base64.media_type: Input should be \'image/jpeg\', \'image/png\', \'image/gif\' or \'image/webp\'"},"request_id":"req_011CVT767kQhcPNzqorczA5M"}No fallback model group found for original model_group=claude-sonnet-4-5-20250929. Fallbacks=[{\'qwen3-coder-480b\': [\'qwen3-coder-480b-or\']}, {\'glm-4.5\': [\'glm-4.5-or\']}]. Received Model Group=claude-sonnet-4-5-20250929\nAvailable Model Group Fallbacks=None\nError doing the fallback: litellm.BadRequestError: AnthropicException - {"type":"error","error":{"type":"invalid_request_error","message":"messages.0.content.0.image.source.base64.media_type: Input should be \'image/jpeg\', \'image/png\', \'image/gif\' or \'image/webp\'"},"request_id":"req_011CVT767kQhcPNzqorczA5M"}No fallback model group found for original model_group=claude-sonnet-4-5-20250929. Fallbacks=[{\'qwen3-coder-480b\': [\'qwen3-coder-480b-or\']}, {\'glm-4.5\': [\'glm-4.5-or\']}] LiteLLM Retried: 1 times, LiteLLM Max Retries: 2', 'type': None, 'param': None, 'code': '400'}}
```

We should only convert it to one of these four types ` \'image/jpeg\', \'image/png\', \'image/gif\' or \'image/webp\'"`

If it failed, we should return text message and fallback to save to disk blob

xingyaoww: @OpenHands if it is image, can we use some library to try to convert them to a compatible format where possible instead of just fall back to the old path

openhands_cli/acp_impl/main.py
```diff
@@ -0,0 +1,60 @@
...
+import os
+import sys
+from datetime import datetime
+
+from openhands.sdk.logger import DEBUG
+
+from .agent import run_acp_server
+
+
+def setup_logging_and_redirect() -> None:
```
xingyaoww: @OpenHands this implementation breaks the stdin/out of the normal ACP:

You can reproduce by running this `./build.sh --install-pyinstaller` and it will fail due to timeout.

I want you to redirect stdin/out OTHER THAN ACP server inputs/outputs

openhands_cli/acp_impl/utils/resources.py
xingyaoww: @OpenHands if it is image, can we use some library to try to convert them to a compatible format where possible instead of just fall back to the old path

xingyaoww:
@neubig screenshot uploading is fixed in this commit: https://github.com/OpenHands/OpenHands-CLI/pull/69/commits/7faa03006e3371afdeda05269242aabe05b5bcb0

xingyaoww:
@OpenHands please fix the failing actions on PR #69 at branch `xw/acp-simplification`

Please revert the stdout/stderr surpression commit if that fixes the issue

hooks/rthook_filter_syspath.py
```diff
@@ -0,0 +1,16 @@
+"""
```
malhotra5: this seems like a openhands cloud specific workaround that the agent made?

openhands_cli/acp_impl/utils/convert.py
```diff
@@ -0,0 +1,97 @@
...
+        elif isinstance(block, ACPImageContentBlock):
+            # Only pass supported image types to ImageContent
+            # Try to convert unsupported types, fall back to disk storage if that fails
+            if block.mimeType in SUPPORTED_IMAGE_MIME_TYPES:
+                message_content.append(
+                    ImageContent(
+                        image_urls=[f"data:{block.mimeType};base64,{block.data}"]
+                    )
+                )
+            else:
```
malhotra5: NIT: could we simplify the nesting of conditionals here

hooks/rthook_filter_syspath.py
xingyaoww: yep, removed

openhands-cli.spec
```diff
@@ -59,8 +62,7 @@ a = Analysis(
     ],
     hookspath=[],
     hooksconfig={},
-    runtime_hooks=[],
-    # runtime_hooks=[str(project_root / "hooks" / "rthook_profile_imports.py")],
+    runtime_hooks=[str(project_root / "hooks" / "rthook_filter_syspath.py")],
```
malhotra5:
```suggestion
    runtime_hooks=[],
    # runtime_hooks=[str(project_root / "hooks" / "rthook_profile_imports.py")],
```

Build is failing because of broken profiler, lets revert that for now

openhands_cli/acp_impl/utils/convert.py
xingyaoww: @OpenHands please simplify the nested conditionals here

Also make sure we have sufficient tests in tests/acp/test_convert.py to cover all branches

openhands-cli.spec
xingyaoww: fixed

scripts/acp/debug_client.py
```diff
@@ -0,0 +1,357 @@
+#!/usr/bin/env python3
```
malhotra5, line 1: what's the purpose of this file as opposed to `scripts/jsonrpc_cli.py` ? 

NIT: let's move this to the `scripts` directory

openhands_cli/acp_impl/event.py
```diff
@@ -0,0 +1,523 @@
...
+                #         sessionId=self.session_id,
+                #         update=UserMessageChunk(
+                #             sessionUpdate="user_message_chunk",
+                #             content=TextContentBlock(
+                #                 type="text",
+                #                 text=viz_text,
+                #             ),
+                #         ),
+                #     )
+                # )
```
malhotra5:
```suggestion
```

openhands_cli/acp_impl/event.py
```diff
@@ -0,0 +1,523 @@
...
+                    # Convert TaskItems to PlanEntries
+                    entries: list[PlanEntry] = []
+                    for task in observation.task_list:
+                        # Map status: todo→pending, in_progress→in_progress,
+                        # done→completed
+                        status_map: dict[str, PlanEntryStatus] = {
+                            "todo": "pending",
+                            "in_progress": "in_progress",
+                            "done": "completed",
+                        }
```
malhotra5: NIT: lets make these enums on our side?

openhands_cli/acp_impl/event.py
```diff
@@ -0,0 +1,523 @@
...
+        elif isinstance(event, SystemPromptEvent):
+            await self._handle_system_prompt_event(event)
+        elif isinstance(event, PauseEvent):
+            await self._handle_pause_event(event)
+        elif isinstance(event, Condensation):
+            await self._handle_condensation_event(event)
+        elif isinstance(event, CondensationRequest):
+            await self._handle_condensation_request_event(event)
+
+    async def _handle_action_event(self, event: ActionEvent):
```
malhotra5, line 150: how does this interact with parallel tool calls? does the client invoke this method on one tool call at a time?

scripts/acp/debug_client.py
xingyaoww: fixed

openhands_cli/acp_impl/event.py
xingyaoww: I think these are the types of "status" of task tracker in SDK, `.venv/lib/python3.12/site-packages/openhands/tools/task_tracker/definition.py`, probably not a good idea to define another Enum inside CLI. 
@OpenHands please update `status: Literal["todo", "in_progress", "done"]` to a `StatusType= Literal[XXX]` task_tracker/definition.py in OpenHands/software-agent-sdk repo, and then update the current PR to improt this type, and change the status map to `dict[TaskTrackerStatusType, PlanEntryStatus]`

openhands_cli/acp_impl/event.py
xingyaoww: > does the client invoke this method on one tool call at a time?

Yes, it will behave the exact same way as CLI / SDK (one tool call at a time). I've been observing this in Zed editor: in practice, it will emit multiple actions at the same time inside IDE

openhands_cli/acp_impl/main.py
```diff
@@ -0,0 +1,26 @@
...
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+    handlers=[logging.StreamHandler(sys.stderr)],
+)
+
+logger = logging.getLogger(__name__)
+
+
+async def run_acp_agent() -> None:
```
malhotra5: NIT: do we need this function wrap, or can we use `run_acp_server` interchangeably ?

openhands_cli/acp_impl/test_utils.py
```diff
@@ -0,0 +1,174 @@
+"""
```
malhotra5, line 1: NIT: let's move this to scripts if its for debugging purposes

openhands_cli/acp_impl/agent.py
```diff
@@ -0,0 +1,477 @@
...
+                        f"Conversation thread did not stop within timeout for session "
+                        f"{params.sessionId}"
+                    )
+                    # Cancel the task if it doesn't stop in time
+                    running_task.cancel()
+                    try:
+                        await running_task
+                    except asyncio.CancelledError:
+                        pass
+                except Exception as e:
```
malhotra5: NIT: should this raise a ACP `RequestError`?

openhands_cli/acp_impl/agent.py
```diff
@@ -0,0 +1,477 @@
...
+            )
+            raise RequestError.internal_error(
+                {"reason": "Failed to process prompt", "details": str(e)}
+            )
+
+    async def cancel(self, params: CancelNotification) -> None:
+        """Cancel the current operation."""
+        logger.info(f"Cancel requested for session: {params.sessionId}")
+
+        try:
```
malhotra5, line 365: NIT: any chance we can simplify the nesting here?

openhands_cli/acp_impl/main.py
xingyaoww: @OpenHands please investigate, if we don't need this, simplify it

openhands_cli/acp_impl/test_utils.py
xingyaoww: We need these to test the executable binary during `build.py`, maybe it is a better idea to put it here?

openhands_cli/acp_impl/agent.py
xingyaoww: @OpenHands please investigate, if so, help me modify code.

openhands_cli/acp_impl/agent.py
xingyaoww: @OpenHands help me simplify the nesting

openhands_cli/acp_impl/event.py
xingyaoww: @OpenHands the SDK changes has just been published to openhands-sdk and tools at version 1.3.0.

Please import TaskTrackerStatusType from it and address the comment here

xingyaoww:
@OpenHands the SDK changes has just been published to openhands-sdk and tools at version 1.3.0.

Please import TaskTrackerStatusType from it and address the comment here: https://github.com/OpenHands/OpenHands-CLI/pull/69#discussion_r2560703265

openhands_cli/acp_impl/agent.py
xingyaoww: done

openhands_cli/acp_impl/main.py
xingyaoww: done

openhands_cli/acp_impl/agent.py
xingyaoww: done

openhands_cli/acp_impl/agent.py
```diff
@@ -0,0 +1,483 @@
...
+
+            # Send the message with potentially multiple content types
+            # (text + images)
+            message = Message(role="user", content=message_content)
+            conversation.send_message(message)
+
+            # Run the conversation asynchronously
+            # Callbacks are already set up when conversation was created
+            # Track the running task so cancel() can wait for proper cleanup
+            run_task = asyncio.create_task(asyncio.to_thread(conversation.run))
```
malhotra5, line 306: QQ: what happens if an error occurs during this run? Is there a callback that gets sent via ACP?

For example, ran out of llm credits, etc

build.sh
```diff
@@ -44,5 +44,9 @@ if [ "$INSTALL_PYINSTALLER" = true ]; then
     fi
 fi
 
+# Clean PYTHONPATH to avoid conflicts with /openhands/code
+# This prevents namespace conflicts between the OpenHands main repo and openhands-sdk packages
+export PYTHONPATH=""
+
```
malhotra5:
```suggestion
```

openhands_cli/acp_impl/agent.py
xingyaoww: I think an exception will be thrown, and it turns into `RequestError.internal_error` :)

xingyaoww:
@malhotra5 Dev guide added, and the docs is ready at https://github.com/OpenHands/docs/pull/116

xingyaoww:
@OpenHands please write a new github issue based on the PR description of this PR to track the need that we will need to implement slash command and confirmation mode (include links to ACP spec)

PR #74, opened 2025-11-18
Fix model list display to show up to 10 models per provider

xingyaoww:
Model display still works!

Before fix:
image

After fix:
image


malhotra5 (approved):
Thank you!

PR #75, opened 2025-11-18
Release 1.2.0 + 1.2.1

malhotra5:


pyproject.toml
```diff
@@ -4,7 +4,7 @@ requires = [ "hatchling>=1.25" ]
 
 [project]
 name = "openhands"
-version = "1.0.6"
+version = "1.2.1"
```
xingyaoww, line 7: Should it be 1.2.0? If we want to make it 1.2.1, we should probably rename the PR

malhotra5: renamed the PR - i made an accidental error so i cut two releases from this PR (1.2.0 and 1.2.1) the tags are on different commit shas

PR #115, opened 2025-11-20
cli: add --task/-t and --file/-f for initial user message

enyst:
This PR adds initial task/file flags for the CLI:

- --task, -t: send provided text as the first user message in a new conversation
- --file, -f: read UTF-8 file contents and send as the first user message

Behavior:
- When resuming (--resume), flags are ignored with a warning to preserve existing conversation state
- On new conversations, the initial message is injected before the first prompt so the TUI/event flow shows it as the initial user message and the agent responds immediately

Help text updated with examples.

All tests and lint pass locally.

### DIFF
--- a/openhands_cli/agent_chat.py
+++ b/openhands_cli/agent_chat.py
@@ -59,7 +59,11 @@ def _print_exit_hint(conversation_id: str) -> None:
     )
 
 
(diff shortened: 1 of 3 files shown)


enyst:
Redundant with https://github.com/OpenHands/OpenHands-CLI/pull/131

PR #113, opened 2025-11-19
feat: Add --version/-v flag support to print CLI version

xingyaoww:
## Summary
This PR implements support for `--version` and `-v` flags in the OpenHands CLI to display the current version number.

## Changes
- Added `__version__` import to `openhands_cli/argparsers/main_parser.py`
- Implemented `--version/-v` argument with argparse's `version` action
- Added comprehensive tests in `tests/test_main.py` for both flag variants
- Tests validate that:
  - Both `--version` and `-v` flags work correctly
  - Exit code is 0 (success)
  - Output contains "OpenHands CLI" and a semantic version number (X.Y.Z format)

## Testing
All tests pass, including:
- ✅ New version flag tests (2 parametrized tests)
- ✅ All existing tests (17 tests total)
- ✅ Linting checks (ruff format, ruff lint, pycodestyle, pyright)

## Example Output
```bash
$ openhands --version
OpenHands CLI 1.2.1

$ openhands -v
OpenHands CLI 1.2.1
```

Fixes #112

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/7966efc6c75845d8bd3cd1f032c2d7cf)

PR #116, opened 2025-11-20
ci: Add binary --version test to CI workflow

xingyaoww:
## Summary
This PR adds a test step to the "CLI - Build binary and optionally release" workflow that verifies the built binary executable can successfully run `openhands --version`.

## Changes
- Added a new CI step "Test binary --version flag" after binary build verification
- The test runs `openhands --version` on the compiled binary
- Validates that:
  - The binary executable runs successfully (exit code 0)
  - Output contains "OpenHands CLI"
  - Output contains a valid semantic version number (X.Y.Z format)
- If any validation fails, the CI stops and prevents uploading broken binaries

## Motivation
Following up on PR #113 which added `--version/-v` flag support, this ensures the feature works correctly in the compiled binary executable. The test acts as a quick smoke test before artifacts are uploaded, preventing distribution of broken binaries.

## Testing
The test will run automatically on both Linux and macOS builds in the CI workflow.

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/37ca589767cd40cca0e926add894a712)

### DIFF
--- a/.github/workflows/cli-build-binary-and-optionally-release.yml
+++ b/.github/workflows/cli-build-binary-and-optionally-release.yml
@@ -76,50 +76,6 @@ jobs:
                   fi
                   echo "✅ Found binaries to upload."
 
-            - name: Test binary --version flag
-              run: |
-                  echo "🧪 Testing binary --version flag..."
-
-                  # Find the binary executable
-                  if [ -f dist/openhands ]; then
-                    BINARY="dist/openhands"
-                  elif [ -f dist/openhands.exe ]; then
-                    BINARY="dist/openhands.exe"
-                  else
-                    echo "❌ Binary executable not found!"
-                    exit 1
-                  fi
-
-                  # Make binary executable on Unix-like systems
-                  if [ "${{ matrix.platform }}" != "windows" ]; then
-                    chmod +x "$BINARY"
-                  fi
-
-                  # Run --version and capture output
-                  echo "Running: $BINARY --version"
-                  if ! VERSION_OUTPUT=$("$BINARY" --version 2>&1); then
-                    echo "❌ Failed to run binary --version command!"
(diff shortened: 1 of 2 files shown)


xingyaoww:
@OpenHands please fix the failing actions on PR #116 at branch `openhands/add-version-test-to-ci`

xingyaoww:
@malhotra5 i actually didn't see that file in this repo 👀 could you point me to the correct file?

yeah i created this test bc i was paranoid, wondering if __version__ will be accessible from CI. It seems so - if you already have a test script, we can just move the content of this PR there; if the test script is not yet ported over to the new repo, i think we can merge this as is and combine them when the tests are being ported.

malhotra5:
Ahh I see the file locations and names aren't very obvious 

Yes the tests are already in this repo, see 

1. https://github.com/OpenHands/OpenHands-CLI/blob/main/build.py (we should rename this. it builds and tests the executable)
2. https://github.com/OpenHands/OpenHands-CLI/blob/main/build.sh (the entrypoint for running the build.py script. we can probably get rid of this file its for convenience. we can move the command in this script to the [MakeFile](https://github.com/OpenHands/OpenHands-CLI/blob/main/Makefile) instead)


I considered doing smoke tests instead of the E2E test every time, but found historically that the smoke test would pass while the executable is actually broken. The E2E was more reliably at catching breaks to the executable (because we lazy load packages and the installer may break deeper into the runtime).

xingyaoww:
@OpenHands please review diff of this PR and this comment https://agentclientprotocol.com/protocol/overview

Move the --version test from CI workflow to the build.py

malhotra5 (changes requested):
Requesting change for now as it seems we're changing the direction of this PR

malhotra5 (approved):

PR #131, opened 2025-11-24
feat(cli): add task & file params

jwerner08:
# Purpose

I'd like to mess around more with the Openhands VS Code extension. I saw that it's currently set up to accept `--task` and `--file` parameters, which the CLI doesn't currently accept. I then found that [this is an existing open issue](https://github.com/OpenHands/OpenHands-CLI/issues/62), so I took a stab at implementing a solution to unblock this workflow. 😸 

I tried to keep my implementation as consistent as possible with existing code patterns and naming idioms, but I’m absolutely open to refactors or style tweaks! 

## What this changes

1. Updates the main argparse setup to accept:

- `--task` / `-t`: a string used to seed the initial conversation message.
- `--file` / `-f`: a path to a file whose contents are used to seed the initial conversation.

2. Wires those arguments into the CLI entrypoint so that:

- An `initial_message` is passed to seed the conversation with.
- `--file` takes precedence over `--task` when both are provided.
- `serve` behavior is unchanged.

3. Adds tests around:

- Parsing of `--task` / `--file`.
- The interaction between those flags (including `--file` taking precedence).

## Before

image

## After

image

image

### DIFF
--- a/openhands_cli/agent_chat.py
+++ b/openhands_cli/agent_chat.py
@@ -61,7 +61,7 @@ def _print_exit_hint(conversation_id: str) -> None:
 
 def run_cli_entry(
     resume_conversation_id: str | None = None,
-    initial_message: str | None = None
+    queued_inputs: list[str] | None = None,
 ) -> None:
     """Run the agent chat session using the agent SDK.
 
@@ -71,6 +71,9 @@ def run_cli_entry(
         KeyboardInterrupt: If user interrupts the session
         EOFError: If EOF is encountered
     """
+
+    # Normalize queued_inputs to a local copy to prevent mutating the caller's list
+    pending_inputs = list(queued_inputs) if queued_inputs else []
 
     conversation_id = uuid.uuid4()
     if resume_conversation_id:
@@ -104,27 +107,17 @@ def run_cli_entry(
     conversation = None
     session = get_session_prompter()
 
-    # Seed initial message (from --task/--file) BEFORE interactive loop
-    if initial_message:
-        if not runner or not conversation:
-            conversation = setup_conversation(conversation_id)
-            runner = ConversationRunner(conversation)
-
-        seed_message = Message(
-            role="user",
-            content=[TextContent(text=initial_message)],
-        )
-        runner.process_message(seed_message)
-        print()
-
     # Main chat loop
     while True:
         try:
             # Get user input
-            user_input = session.prompt(
-                HTML("<gold>> </gold>"),
-                multiline=False,
-            )
+            if pending_inputs:
+                user_input = pending_inputs.pop(0)
+            else:
+                user_input = session.prompt(
+                    HTML("<gold>> </gold>"),
+                    multiline=False,
+                )
 
             if not user_input.strip():
                 continue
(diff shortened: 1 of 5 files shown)


openhands_cli/argparsers/main_parser.py
```diff
@@ -15,15 +15,15 @@ def create_main_parser() -> argparse.ArgumentParser:
...
-""",
+            By default, OpenHands runs in CLI mode (terminal interface).
+            Use 'serve' subcommand to launch the GUI server instead.
+
+            Examples:
+                openhands                           # Start CLI mode
+                openhands --resume conversation-id  # Resume a conversation in CLI mode
+                openhands serve                     # Launch GUI server
+                openhands serve --gpu               # Launch GUI server with GPU support
+        """,
```
jwerner08, line 26: This is just a formatting change cuz it hurt my eyes 😵‍💫

README.md
jwerner08, line 1: This change was made by my markdown formatter that I have set to format on save. I can remove if needed, but I think it makes the readme easier to read, even though it's just whitespace.

openhands_cli/agent_chat.py
```diff
@@ -101,6 +104,19 @@ def run_cli_entry(resume_conversation_id: str | None = None) -> None:
...
+        if not runner or not conversation:
+            conversation = setup_conversation(conversation_id)
+            runner = ConversationRunner(conversation)
+
+        seed_message = Message(
+            role="user",
+            content=[TextContent(text=initial_message)],
+        )
+        runner.process_message(seed_message)
+        print()
```
malhotra5: I'm thinking we rework this as following

1. instead of `initial_message: str`, lets make the param to this function `queued_inputs: list[str]`. For the scope of this PR, `queued_inputs` will contain exactly one seed instruction

2. instead of duplicating the `process_message` call here, we can modify the main chat loop below as follows

```
if queued_inputs:
    user_input = queued_inputs.pop(0)
else:
    user_input = session.prompt(...)
```

then the chat loop will handle the instructions as it does normally

This gives us two benefits

1. in the future we can extend this seeding functionality to process a list of instructions sequentially
2. messages can also serve as commands. for example, if the seed instruction was `\exit` then the CLI would exit since the main chat loop is handling this logic. this would greatly simplify E2E tests we have as well


LMK if this seems good!

openhands_cli/simple_main.py
```diff
@@ -30,6 +31,30 @@ def main() -> None:
...
+            initial_message = (
+                "Starting this session with file context.\n\n"
+                f"File path: {path}\n\n"
+                "File contents:\n"
+                "--------------------\n"
+                f"{content}\n"
+                "--------------------\n"
+            )
+        elif getattr(args, "task", None):
+            initial_message = args.task
```
malhotra5: NIT: lets move this logic into a util file

openhands_cli/agent_chat.py
jwerner08: Generally I try avoiding abstractions until they're forced, but I do see immediate value here since it's more idiomatic and keeps command handling in the main loop. 👍 

Will rework this!

jwerner08:
Hey @malhotra5! 👋 

Addressed your feedback, so this PR is ready for a re-review whenever you have time. 

Thank you!

malhotra5 (approved):
LGTM! Thanks you!

PR #133, opened 2025-11-25
Restrict Python version to <3.14 to avoid tiktoken build issues

xingyaoww:
## Summary

This PR adds an upper bound to the `requires-python` constraint in `pyproject.toml` to exclude Python 3.14, preventing installation failures.

**Change:**
```diff
- requires-python = ">=3.12"
+ requires-python = ">=3.12,<3.14"
```

## Problem

Users were experiencing installation failures when running `uv tool install openhands` because:
1. `uv` was selecting Python 3.14 (if available on the system)
2. `tiktoken==0.11.0` and potentially other dependencies lack prebuilt wheels for Python 3.14
3. This caused compilation from source, requiring rust compiler installation
4. Users encountered errors like:
   ```
   × Failed to build tiktoken==0.11.0
   ```

Reference: User reports in Slack where multiple users couldn't install the CLI using `uv`, and logs showing installation attempts with Python 3.14.

## Solution

By adding an upper bound to `requires-python`, we leverage uv's automatic Python version selection to:
- ✅ Allow Python 3.12 and 3.13 (which have proper wheel support)
- ✅ Automatically exclude Python 3.14 (which causes build issues)
- ✅ Work seamlessly without users needing to specify `--python 3.12` flag
- ✅ Align with the project's supported versions (3.12, 3.13 as listed in classifiers)

According to [uv's documentation on Python versions](https://docs.astral.sh/uv/concepts/python-versions/#project-python-versions), `requires-python` in `pyproject.toml` is respected during tool installation, ensuring uv selects a compatible Python version automatically.

## Benefits

- Users can simply run `uv tool install openhands` without additional flags
- Installation will work reliably across different environments
- No unexpected compilation requirements or rust compiler errors
- Cleaner, more maintainable solution than documenting workarounds

## Testing

This change will be validated when:
- Users run `uv tool install openhands` on systems with Python 3.14
- uv should automatically select Python 3.12 or 3.13 instead
- Installation should complete successfully without build errors

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/413800b275e24744a4de26a1ba7657ff)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ description = "OpenHands CLI - Terminal User Interface for OpenHands AI Agent"
 readme = "README.md"
 license = { text = "MIT" }
 authors = [ { name = "OpenHands Team", email = "contact@all-hands.dev" } ]
-requires-python = ">=3.12,<3.14"
+requires-python = "==3.12.*"
 classifiers = [
   "Programming Language :: Python :: 3 :: Only",
   "Programming Language :: Python :: 3.12",


enyst (approved):
Thank you! This doesn't allow 3.13 either, and IMHO that is maybe the right choice for now.

To note, lately I haven't heard of 3.13 issues. On the other hand, I don't think we've tested it... not even semi-officially? So 3.12.x seems safe to me.

PR #138, opened 2025-11-26
Release 1.3.0

malhotra5:


uv.lock
```diff
@@ -1,10 +1,6 @@
 version = 1
 revision = 3
-requires-python = ">=3.12"
-resolution-markers = [
-    "python_full_version >= '3.13'",
-    "python_full_version < '3.13'",
-]
+requires-python = "==3.12.*"
```
enyst, line 3: Wow it removed a lot of package versions. I think that was expected?

enyst (approved):

uv.lock
malhotra5: Yep! I believe so

PR #162, opened 2025-12-04
Load .env at CLI startup to avoid python-dotenv stack-walk failures

enyst:
Summary

Load .env once at CLI startup to make configuration deterministic and avoid python-dotenv's find_dotenv() call-stack introspection path that can raise AssertionError in wrapped/decorated execution contexts.

What changed
- openhands_cli/simple_main.py: load .env from the current working directory if it exists, using load_dotenv(dotenv_path=..., override=False)

Why
- We observed AssertionError from python-dotenv's find_dotenv() when invoked indirectly deep in execution (see software-agent-sdk#1325). Loading dotenv at startup avoids stack walking and repeated parsing.
- This centralizes .env handling in the application layer (CLI), aligning with best practices that SDKs should be side-effect free on import.

Behavior
- If a .env file exists in CWD, its variables are loaded into os.environ once at startup. Values already in the environment are not overridden.
- If no .env file is present, nothing happens.

Tests
- pre-commit (ruff/pyright/pycodestyle) passes on the modified file
- `uv run pytest` passes locally for the CLI repo

Notes
- This is an additive, non-breaking change. Application-controlled precedence is preserved thanks to override=False.

---

---

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@chore/load-dotenv-at-startup
```

### DIFF
--- a/openhands_cli/simple_main.py
+++ b/openhands_cli/simple_main.py
@@ -7,7 +7,6 @@ This is a simplified version that demonstrates the TUI functionality.
 import logging
 import os
 import warnings
-
 from pathlib import Path
 
 from dotenv import load_dotenv


openhands_cli/simple_main.py
```diff
@@ -8,6 +8,10 @@
 import os
 import warnings
 
+# Load environment variables from .env early, if present
```
enyst:
```suggestion
```

enyst:
GPT-5 argued that it's unexpected to load it in an SDK/library 🤔 

> **Short answer:** No—SDKs should not auto-load dotenv at import/initialization time.
> 
> **Why**
> - Principle of least surprise: Libraries should avoid side effects on import. Auto-loading .env mutates process state and can conflict with host apps’ configuration order.
> - Ownership: The application (CLI/server) owns configuration and should decide when/how to load .env, including precedence (override or not), path, and environment specifics.
> - Compatibility: Some hosts may manage secrets via env vars, containers, or orchestrators; silently reading .env in the SDK can introduce hard-to-trace behavior.
> 
> **Recommended pattern**
> - Application entry point: Call load_dotenv(usecwd=True, override=False) early. This avoids python-dotenv’s fragile stack-walking and sets env once.
> - SDK behavior: Keep get_env reading os.environ first. If we keep a dotenv fallback, use a safe path resolution (e.g., find_dotenv(usecwd=True, raise_error_if_not_found=False) with defensive try/except) rather than relying on find_dotenv’s call-stack inspection.
> - Optional: Provide an explicit, opt-in helper (e.g., openhands.sdk.init_config(load_dotenv=True, override=False)) for apps that prefer the SDK to manage this, but keep default behavior side-effect free.

enyst:
I'll merge this, I think it's correct, even though we do some and IMO we should continue to do some (configurable) loading in the SDK.

PR #141, opened 2025-11-27
Refactor CLI UI 

malhotra5:
This PR uses the textual library to refactor our CLI UI

This PR should address the following issues

- [X] Spinner or seconds counter for long running agent commands #117
- [X] Queue prompts #103
- [X] Search files with the `@` macro #123
- [X] Sets the groundwork for #124 (although the current UI will override the light terminal color scheme, which is an alternate yet acceptable temporary workaround for now)
- [X] Reduce verbosity of messages and make them collapsible [ALL-4117](https://linear.app/all-hands-ai/issue/ALL-4117/cli-reduce-verbosity-of-cli-messages)

Required feature parity

- [x] CLI args handling
- [x] Security analyzer
- [x] MCP screen
- [x] Settings screen
- [x] Pausing conversation
- [X] Running agent loop
- [x] Multi-line input

In progress items to fix

- [x] Better lower verbosity messages
- [x] Coloring on collapsible message events
- [x] Sidebar panel for confirmation mode
- [x] Footer for existing macros
- [x] Fix message copy contents
- [x] Autoscroll for events 

You can run the CLI on this branch with `--exp` flag to try out the new UI. This PR does not remove any existing functionality or UI from the CLI and introduces the new changes behind a feature flag. 

We do this because 
1. the pr is very large 
2. we can continue polishing the new UI and flows before we're confident making it the default experience

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@refactor-ui
```

PR #170, opened 2025-12-05
Fix MarkupError with Chinese characters and special symbols

xingyaoww:
## Problem

The CLI was experiencing a `MarkupError` when rendering content containing Chinese characters combined with special markup characters like `[`, `]`, `+`, and `%`.

Example error:
```
MarkupError: Expected markup value (found '+0.3%,月变化+0.8%,处于历史40%分位]\n').
```

## Root Cause

When `event.visualize` (a Rich renderable object) is converted to a string and passed to `Static` widgets with `markup=True` (the default), the Static widget attempts to parse it as Rich markup. However, when the content contains literal square brackets `[` and `]` (not meant as markup tags), combined with Chinese characters and other special symbols, the markup parser fails.

## Solution

This PR adds proper escaping of Rich markup characters before content is passed to Static widgets:

1. **New helper method**: `_safe_content_string()` that escapes square brackets in content strings to prevent them from being interpreted as markup tags
2. **Applied consistently**: All instances where `str(content)` was used have been replaced with `self._safe_content_string(content)` in the visualizer
3. **Preserves formatting**: Intentional Rich markup (like colored metrics with `[cyan]`, `[magenta]`, etc.) continues to work because it's added after the content is escaped

## Changes

- Modified `openhands_cli/refactor/widgets/richlog_visualizer.py`:
  - Added `_safe_content_string()` method to escape markup characters
  - Updated all event handlers (ActionEvent, ObservationEvent, MessageEvent, AgentErrorEvent, PauseEvent, Condensation, etc.) to use the safe string conversion

## Testing

- [x] Linting passes (`make lint`)
- The fix ensures content with Chinese characters and special symbols can be displayed without markup errors
- Intentional Rich markup in metrics displays continues to function correctly

## Related Issue

Fixes the Chinese character MarkupError reported in PR #141

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/b486f1a5d4554064b688873dafb38d7b)

### DIFF
--- a/tests/refactor/widgets/test_richlog_visualizer.py
+++ b/tests/refactor/widgets/test_richlog_visualizer.py
@@ -1,4 +1,4 @@
-"""Tests for TextualVisualizer and Chinese character markup handling."""
+"""Tests for ConversationVisualizer and Chinese character markup handling."""
 
 from typing import TYPE_CHECKING
 
@@ -12,7 +12,7 @@ from textual.widgets import Static
 from openhands.sdk.event import ActionEvent, MessageEvent
 from openhands.sdk.llm import MessageToolCall, TextContent
 from openhands.sdk.tool import Action
-from openhands_cli.refactor.widgets.richlog_visualizer import TextualVisualizer
+from openhands_cli.refactor.widgets.richlog_visualizer import ConversationVisualizer
 
 
 if TYPE_CHECKING:
@@ -45,7 +45,7 @@ class TestChineseCharacterMarkupHandling:
         # Create a mock app and container for the visualizer
         app = App()
         container = VerticalScroll()
-        visualizer = TextualVisualizer(container, app)  # type: ignore[arg-type]
+        visualizer = ConversationVisualizer(container, app)  # type: ignore[arg-type]
 
         # Test escaping with various bracket patterns
         test_cases = [
@@ -69,7 +69,7 @@ class TestChineseCharacterMarkupHandling:
         """Test that _escape_rich_markup escapes MarkupError content."""
         app = App()
         container = VerticalScroll()
-        visualizer = TextualVisualizer(container, app)  # type: ignore[arg-type]
+        visualizer = ConversationVisualizer(container, app)  # type: ignore[arg-type]
 
         # Example content that caused the original error
         problematic_content = "+0.3%,月变化+0.8%,处于历史40%分位]"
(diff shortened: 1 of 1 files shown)


xingyaoww:
@OpenHands please add a test and make sure before this fix that test will fail, and after this PR it will success

xingyaoww:
@OpenHands please fix the failing actions on PR #170 at branch `fix-chinese-markup-error`

xingyaoww:
@OpenHands please fix the failing actions on PR #170 at branch `fix-chinese-markup-error`

openhands_cli/refactor/widgets/richlog_visualizer.py
```diff
@@ -107,6 +107,15 @@ def _escape_rich_markup(self, text: str) -> str:
...
         return text.replace("[", r"\[").replace("]", r"\]")
 
+    def _safe_content_string(self, content) -> str:
+        """Convert content to string and escape markup characters to prevent errors.
+
+        This is needed to handle content with special characters (e.g., Chinese text
+        with brackets) that would otherwise cause MarkupError when rendered in Static
+        widgets with markup=True.
+        """
+        return self._escape_rich_markup(str(content))
```
xingyaoww: @OpenHands can we just reuse `_escape_rich_markup` instead of creating another function here?

xingyaoww:
@OpenHands please fix the merge conflict

malhotra5:
seems to be a merge conflict on the PR

openhands_cli/refactor/widgets/richlog_visualizer.py
```diff
@@ -247,15 +252,15 @@ def _create_event_collapsible(self, event: Event) -> NonClickableCollapsible | N
         elif isinstance(event, ObservationEvent):
             title = self._extract_meaningful_title(event, "Observation")
             return NonClickableCollapsible(
```
malhotra5, line 254: I think that in the constructor of this class we can add `markup=False` to `super().__init__()` - it should behave the same

we can dedup lots of code that way

openhands_cli/refactor/widgets/autocomplete.py
```diff
@@ -18,6 +18,8 @@ class EnhancedAutoComplete(AutoComplete):
     def __init__(self, target, command_candidates=None, **kwargs):
         """Initialize with command candidates and no static candidates."""
         self.command_candidates = command_candidates or []
+        # Remove 'candidates' from kwargs if present to avoid duplicate argument
+        kwargs.pop("candidates", None)
```
malhotra5: is this needed?

xingyaoww: @OpenHands please remove it

openhands_cli/refactor/widgets/richlog_visualizer.py
malhotra5: @openhands figure out where this is possible as opposed to repeatedly calling `self._escape_rich_markup` 


if possible, make the change and push them. ensure that the tests in `tests/refactor/widgets/test_richlog_visualizer.py` are still passing

PR #188, opened 2025-12-09
Fix error display crash when error message contains XML special characters

jpshackelford:
## Problem

When an error occurs that contains XML special characters (like `<`, `>`, `&`), the error display handler crashes with `xml.parsers.expat.ExpatError` because prompt_toolkit's `HTML()` formatter tries to parse the raw error message as XML.

This was seen when the agent state validation failed with a diff containing `'<missing>'` and `->` characters.

## Solution

Use `html.escape()` to properly escape XML special characters before passing the error message to the HTML formatter.

## Changes

- Added `html.escape()` call in the exception handler in `simple_main.py` (line 108)

## Testing

Added comprehensive test suite in `tests/test_error_display.py` with 4 tests covering:
- XML special characters (`<`, `>`, `->`, `{}`)
- Angle brackets
- Ampersands
- Quotes

### Test validates the fix

**Without the fix** (html.escape commented out):
```
FAILED - xml.parsers.expat.ExpatError: mismatched tag: line 7, column 81
```

**With the fix**:
```
4 passed, 2 warnings in 0.22s ✓
```

## Impact

- ✅ Error messages with special characters are now displayed correctly
- ✅ No more secondary crashes when displaying error messages
- ✅ Users can see the actual error instead of ExpatError

---

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@jps/fix-error-display-crash
```

### DIFF
--- a/tests/test_error_display.py
+++ b/tests/test_error_display.py
@@ -8,7 +8,7 @@ import pytest
 
 def test_error_with_xml_special_characters(monkeypatch):
     """Test that error messages with XML special characters are displayed correctly.
-    
+
     This reproduces the bug where error messages containing < > & characters
     would cause xml.parsers.expat.ExpatError when the error handler tried to
     display them using prompt_toolkit's HTML formatting.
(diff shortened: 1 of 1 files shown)


malhotra5 (approved):
Thank you!

PR #194, opened 2025-12-10
Feat: switch models mid conversation

malhotra5:
## Summary

This PR allows users to switch their settings mid conversation and have it take effect immediately. This is done by re-instantiating conversation runner after settings are already saved

In the future, we would employ a similar mechanism for when we have LLM profiles 

---

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@fix/settings-screen-behavior
```


enyst (approved):

malhotra5:
There was a problem reconciling the LLM against persisted state (need to dig into the sdk upstream probably to fix that) - closing this PR for now

PR #205, opened 2025-12-12
fix(cli): add terminal compatibility detection

tosincarik:
## Summary

This PR adds a small, focused terminal compatibility check to the OpenHands CLI, without changing any existing public APIs or unrelated behavior.

The goal is to gracefully handle environments where the interactive TUI may not work correctly (e.g., non-TTY outputs, very limited terminals), while keeping defaults non-breaking and allowing automation to continue to work.

---

## Changes

- Added `openhands_cli.terminal_compat` with:
  - [TerminalCompatibilityResult](cci:2://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/openhands_cli/terminal_compat.py:5:0-9:16) dataclass (`compatible`, `reason`, `is_tty`).
  - [check_terminal_compatibility(stdout=None, env=None)](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/openhands_cli/terminal_compat.py:19:0-52:5):
    - Uses [stdout.isatty()](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/tests/test_terminal_compat.py:13:4-14:27) (if available) to detect TTY.
    - Treats non-TTY stdout as incompatible (TUI may not render correctly).
    - Treats `TERM` unset/empty or `"dumb"` as incompatible, even if TTY.
    - Returns a simple, structured result used by the CLI.
  - [strict_mode_enabled(env=None)](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/openhands_cli/terminal_compat.py:55:0-58:67):
    - Reads `OPENHANDS_CLI_STRICT_TERMINAL` from the environment.
    - Enables strict mode for values like `1`, `true`, `yes`, `on` (case-insensitive).

- Integrated the compatibility check into [run_cli_entry](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/openhands_cli/agent_chat.py:69:0-268:18) in `openhands_cli.agent_chat`:
  - Called **after** agent setup succeeds and **before** `display_welcome` and prompt session creation.
  - On incompatibility:
    - Builds a single explanatory message:
      - `OpenHands CLI terminal UI may not work correctly in this environment: {reason}`
    - If `stdout` is a TTY:
      - Prints as prompt_toolkit `HTML` (yellow warning).
    - If `stdout` is **not** a TTY:
      - Prints using plain `print(message)` to avoid raw HTML/ANSI tags in logs.
    - If [strict_mode_enabled()](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/openhands_cli/terminal_compat.py:55:0-58:67) is `True`:
      - Exits with status code `2`.
    - Otherwise:
      - Continues execution with a warning only (non-strict, non-breaking default).

- Added [tests/test_terminal_compat.py](cci:7://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/tests/test_terminal_compat.py:0:0-0:0):
  - Tests non-TTY behavior: incompatible, `is_tty=False`, reason mentioning “not a TTY”.
  - Tests `TERM="dumb"`: incompatible, `is_tty=True`, reason mentioning “dumb”.
  - Tests a normal terminal (e.g. `TERM="xterm-256color"`): compatible, no reason.
  - Tests [strict_mode_enabled](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/openhands_cli/terminal_compat.py:55:0-58:67) for various true-ish and false-ish env values.

---

## Addressing previous review feedback

This PR is intentionally scoped and avoids the issues raised by the OpenHands bot in the earlier PR:

- **No breaking API changes**
  - [run_cli_entry](cci:1://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/openhands_cli/agent_chat.py:69:0-268:18) signature remains unchanged.
  - `display_welcome` signature and behavior (including version banner) are unchanged.

- **Non-TTY behavior**
  - Incompatibility messages in non-TTY environments are printed as plain text via `print()`, avoiding prompt_toolkit HTML/ANSI noise in logs.

- **No lockfile / Python policy changes**
  - `uv.lock` and Python version markers are untouched in this PR.

- **No additional scope creep**
  - No changes to model identifier sanitization, settings flows, or TUI / version banner behavior.
  - The change set is limited to:
    - [openhands_cli/terminal_compat.py](cci:7://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/openhands_cli/terminal_compat.py:0:0-0:0)
    - [openhands_cli/agent_chat.py](cci:7://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/openhands_cli/agent_chat.py:0:0-0:0) (small integration hook)
    - [tests/test_terminal_compat.py](cci:7://file:///c:/Users/brass/OneDrive/Desktop/Work/OpenHands-CLI/tests/test_terminal_compat.py:0:0-0:0)

- **Strict vs non-strict exit semantics**
  - Default behavior: warn but continue (keeps automation and existing workflows working).
  - Opt-in strict behavior: set `OPENHANDS_CLI_STRICT_TERMINAL` to a true-ish value to make incompatible terminals cause a non-zero exit (`2`).

---

## Usage notes

- In most environments, nothing changes; compatible terminals proceed as before.
- In headless / non-TTY environments:
  - The CLI will emit a **plain-text warning** about potential incompatibility.
  - To enforce a hard failure instead of a warning, set:
    - `OPENHANDS_CLI_STRICT_TERMINAL=true`

### DIFF
--- a/openhands_cli/simple_main.py
+++ b/openhands_cli/simple_main.py
@@ -178,7 +178,6 @@ def main() -> None:
                     "To override Rich's detection, you can set TTY_INTERACTIVE=1 "
                     "(and optionally TTY_COMPATIBLE=1)."
                 )
-                
             # Handle resume logic (including --last and conversation list)
             resume_id = handle_resume_logic(args)
             if resume_id is None and (args.last or args.resume == ""):


malhotra5:
Hi! We have a new experimental UI in the `/refactor` subdirectory that uses the [textual](https://textual.textualize.io/) library that we plan on making the default experience soon


Do you think this UI will also require a similar terminal compatibility check?

tosincarik:
> Hi! We have a new experimental UI in the `/refactor` subdirectory that uses the [textual](https://textual.textualize.io/) library that we plan on making the default experience soon
> 
> Do you think this UI will also require a similar terminal compatibility check?

Thanks for the heads-up — that’s exciting.

Yes, I think the `/refactor` Textual UI will likely benefit from a similar compatibility check, even if the exact failure modes differ from `prompt_toolkit`.

- TTY/non-interactive output still matters: Textual (like most full-screen TUIs) expects to run on a real TTY. If `stdout`/`stdin` isn’t a TTY (CI logs, pipes, `nohup`, some Docker/logging setups), it’s still a common source of “blank screen”, broken rendering, or confusing behavior.
- `TERM=dumb` / minimal terminals still matter: Even with a different renderer, a “dumb” or extremely limited terminal can’t reliably support cursor movement, screen clearing, etc.
- Textual may already have internal detection, but (from a CLI UX perspective) having a small pre-flight check lets us:
   - emit a clear, actionable warning (or fail in strict mode),
   - avoid entering a UI mode that can’t display correctly,
   - keep automation non-breaking by default.

If the plan is to make Textual the default soon, I’d suggest we either:
- reuse the same lightweight pre-check (TTY + `TERM` sanity) before launching the Textual app, or
- wrap/augment Textual’s own detection (if there’s a recommended API), but still keep a single CLI-level “compatibility gate” so behavior is consistent across UIs.

malhotra5:
Thanks for the clarification! And nice catch on textual perhaps exposing a terminal detection helper

I think we can use the [`is_terminal`](https://github.com/Textualize/rich/blob/master/rich/console.py#L937) property from Rich (a library we are already using today, and integrates very nicely with Textual)

This way we can trigger the same env overrides for terminal compatibility (for instance, if we want to run E2E tests in CI that doesn't support TTY we can force compatibility check to pass) + we can dedup a bunch of code

LMK what you think!

tosincarik:
> Hi! We have a new experimental UI in the `/refactor` subdirectory that uses the [textual](https://textual.textualize.io/) library that we plan on making the default experience soon
> 
> Do you think this UI will also require a similar terminal compatibility check?

That makes a lot of sense to me.

Using the [is_terminal](https://github.com/Textualize/rich/blob/master/rich/console.py#L937) property from Rich feels like the right abstraction here, especially since:
- it already encodes the “is this a real interactive terminal?” decision in a way that’s consistent with Rich/Textual,
- it supports env-driven overrides (useful for CI/E2E where TTY isn’t present but we still want UI code paths exercised),
- it avoids duplicating low-level TTY/TERM heuristics in the CLI.

My PR’s intent was mainly:
- warn (or fail in strict mode) when the UI likely won’t render,
- keep logs clean in non-terminal contexts,
- stay non-breaking by default.

tosincarik:
Two quick questions so I align with your preferred semantics:

1. Should we treat `TERM=dumb` as incompatible only insofar as Rich considers it non-terminal, or do you still want an explicit `TERM=dumb` guard even if `is_terminal` is `True`?
2. Do you prefer a dedicated helper like `check_terminal_compatibility(console=...)` (so we reuse the same `Console` config/overrides), or should we instantiate a minimal `Console` internally?

malhotra5:
Re 1: I think we rely completely on Rich. There are some envs that I believe rich detects internally that can be set to force return False (need to double check the accuracy on this though)

Re 2: Probably dedicated helper `check_terminal_compatibility(console=...)`

tosincarik:
@malhotra5 
If you have time, please review this PR for me.

openhands_cli/terminal_compat.py
```diff
@@ -0,0 +1,47 @@
+import os
+from dataclasses import dataclass
+
+from rich.console import Console
+
+
+@dataclass
```
malhotra5: NIT: let's use pydantic basemodel  instead of dataclasses (as we have that convention)

openhands_cli/terminal_compat.py
```diff
@@ -0,0 +1,47 @@
...
+    return TerminalCompatibilityResult(
+        compatible=True,
+        reason=None,
+        is_tty=is_terminal,
+    )
+
+
+def strict_mode_enabled(env: dict[str, str] | None = None) -> bool:
+    if env is None:
+        env = os.environ  # type: ignore[assignment]
```
malhotra5: out of curiosity, we do we have type ignore here?

openhands_cli/agent_chat.py
```diff
@@ -110,20 +115,31 @@ def run_cli_entry(
...
+            "OpenHands CLI terminal UI may not work correctly in this environment: "
+            f"{compat_result.reason}"
+        )
+        if compat_result.is_tty:
+            console.print(message, style="yellow")
+        else:
+            print(message)
+
+        if strict_mode_enabled():
+            sys.exit(2)
```
malhotra5: It seems like this forced the app to exit if true

could we make it so that we're proceeding without exiting in the event we're not `compat_result.compatible`

openhands_cli/terminal_compat.py
```diff
@@ -0,0 +1,47 @@
...
+                "interactive UI may not render correctly"
+            ),
+            is_tty=is_terminal,
+        )
+
+    return TerminalCompatibilityResult(
+        compatible=True,
+        reason=None,
+        is_tty=is_terminal,
+    )
```
malhotra5, line 27: NIT: it seems like `compatible` and `is_tty` are always the same values; maybe we dedup and have just one attribute?

openhands_cli/terminal_compat.py
tosincarik: Good call - I have updated `TerminalCompatibilityResult` to use `pydantic.BaseModel` to match the repo convention. Thanks!

openhands_cli/terminal_compat.py
tosincarik: Makes sense - I deduped this and kept a single boolean (is_tty) + optional reason. Removed the redundant compatible field and updated call sites/tests accordingly.

openhands_cli/terminal_compat.py
tosincarik: the `type: ignore` wasn’t necessary. I removed it by typing `env` as `Mapping[str, str] | None` so `os.environ` is accepted without a cast/ignore.

openhands_cli/agent_chat.py
tosincarik: I changed the behavior so we still warn when the terminal is not compatible, but we continue running (even when `OPENHANDS_CLI_STRICT_TERMINAL` is set). No forced exit anymore.

tosincarik:
> Thanks a bunch we're close! Left some nits
> 
> (resolving merge conflict would be great so we can run the CI!)

I rebased and resolved the conflicts on my branch. The PR branch should now be conflict-free and CI-ready.

tosincarik:
@malhotra5
If you have time, please review this PR for me.

openhands_cli/terminal_compat.py
```diff
@@ -0,0 +1,43 @@
...
+
+    return TerminalCompatibilityResult(
+        is_tty=is_terminal,
+    )
+
+
+def strict_mode_enabled(env: Mapping[str, str] | None = None) -> bool:
+    if env is None:
+        env = os.environ
+    return _env_flag_true(env.get("OPENHANDS_CLI_STRICT_TERMINAL"))
```
malhotra5: I think we may not need this function; it seems like rich supports a way to force compatibility via env `TTY_INTERACTIVE` 

```
if force_interactive is None:
            tty_interactive = self._environ.get("TTY_INTERACTIVE", None)
            if tty_interactive is not None:
                if tty_interactive == "0":
                    force_interactive = False
                elif tty_interactive == "1":
                    force_interactive = True
```


Ideally we can document that somewhere and it should behave the same as this function's intended purpose

openhands_cli/agent_chat.py
```diff
@@ -10,6 +10,7 @@
```
malhotra5: Ah unfortunately this file has been removed. The new CLI UI is now the default entrypoint but can also benefit from this compatibility check

openhands_cli/terminal_compat.py
tosincarik: Updated `terminal_compat.py` to the simplified version requested (no `OPENHANDS_CLI_STRICT_TERMINAL` / `strict_mode_enabled`).

openhands_cli/agent_chat.py
tosincarik: Moved the compatibility check to the new default entrypoint (`simple_main.py`)

tosincarik:
> Really sorry about the recent changes affecting your PR! There seems to be some more room for simplification as well, please feel free to request a review and thank you for reminding me!!

@malhotra5
If you have time, please review this PR for me.

malhotra5:
@tosincarik it seems lint and some tests are failing, mind taking a look? I think its related to outdated functions in previous commits

tosincarik:
> @tosincarik it seems lint and some tests are failing, mind taking a look? I think its related to outdated functions in previous commits

@malhotra5 
I have corrected the failing check

tosincarik:
@malhotra5 
I have corrected the failing check, if you have time, please review this PR for me.

tosincarik:
@malhotra5 
I have corrected the failing check, if you have time, please review this PR for me.

neubig:
Hi @tosincarik , we'll get to this soon, but please give us a few days, given that it's the end of the year and people are generally a bit less available! Also, I still see on failing check about updating the PR description, please take a look.

malhotra5 (approved):
LGTM thank you for the contribution!

PR #218, opened 2025-12-16
fix: path field in confirmation message of FileEditorAction truncating file name

seekayel:
When the agent would ask for confirmation of `FileEditorAction` actions the path to the file was being truncated so it was not clear which file the action was being performed on.

This change modifies the format of the confirmation message in the case of `FileEditorAction` to display a relative path, eg `/Users/ckl/git/seekayel/openhands-github-action-local/README.md` -> `README.md`

## Example before change
```
🔍 Agent created 1 action(s) and is waiting for confirmation:

  1. file_editor: kind='FileEditorAction' command='str_replace' path='/Users/ckl/git/seekayel/openhands-github-action-...

Choose an option:
```

## Example after change
```
🔍 Agent created 1 action(s) and is waiting for confirmation:

  1. file_editor: kind='FileEditorAction' command='str_replace' path='README.md'...

Choose an option:
```

### DIFF
--- a/openhands_cli/user_actions/agent_action.py
+++ b/openhands_cli/user_actions/agent_action.py
@@ -1,5 +1,5 @@
 import html
-
+import os
 from prompt_toolkit import HTML, print_formatted_text
 
 from openhands.sdk.security.confirmation_policy import (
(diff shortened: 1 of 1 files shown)


malhotra5:
Thank you for the contribution @seekayel ! 

We have a new experimental UI in the `/refactor` subdirectory (which we plan on making the default experience soon). You can access it by adding `--exp` flag to however you run the CLI

Is the truncation still an issue in the new UI?

seekayel:
It is. Shall I port this change into there as well?

Screenshot 2025-12-16 at 12 31 12 PM

malhotra5:
That would be greatly appreciated! Feel free to re-request a review from me when thats ready!

seekayel:
@malhotra5 : can you confirm my approach:

seems like the best place to put this change would be in OpenHands/software-agent-sdk repo in the: `openhands-tools/openhands/tools/file_editor/definition.py`

I would modify this method:
https://github.com/OpenHands/software-agent-sdk/blob/97652be0db34010c9f351c7bc8470f9b8a40fe93/openhands-tools/openhands/tools/file_editor/definition.py#L92

I had hesitated to solve an issue in the CLI with a change to the SDK but maybe that is what is called for.

malhotra5:
@seekayel is the sdk truncating paths?

I think it isn't truncating the values, so maybe we would benefit from calling the `.visualize` method in the CLI?

seekayel:
@malhotra5 you are right. `visualize` is already getting called, and its not truncating, just looks like it since it appends the `...`

My bad. Sorry for any confusion 🙏 

Closing since this is handled with new UI refactor.

malhotra5:
@seekayel No worries, thank you for looking into this!

Issue #147, opened 2025-08-20
[PRD] Support non-interactive CLI mode


jpelletier1:
## Problems to solve

* Current Headless mode is not at parity with [OpenHands CLI](https://docs.all-hands.dev/usage/how-to/cli-mode)
  * No MCP support
  * Designed originally for debugging OpenHands; but not for building scalable automation/scripting
* Users want a non-interactive CLI experience with OpenHands, but the current Headless mode is difficult to install, hard to use
  * Hard to configure
  * Requires docker and following development instructions
  * Many users don’t have `tmux` installed

## Customer / User evidence

* Some customers have built custom integrations and interact with OpenHands through the existing websocket connection.
* Several open source users have built custom automation around OpenHands, but have made significant changes or forks to make this possible.
* Interviews with Open Source users have indicated they'd prefer to start with a non-interactive CLI mode instead of the current Terminal UI. These users prefer to build custom automation that allows them to run multiple conversations at once.

## Goals

* easy to install
* ability to run as a command line
* works on Windows, Mac, and Linux environments

## **Requirements**

* Should not require docker (already accomplished with current CLI)
* After a command is entered, display a counter/spinner to show the user the CLI is working
* Allows for users to start conversations from the command line into the Terminal UI
  * `openhands "start a convo"`
* Allows users to run one-off prompts - outputs as text (max verbosity)
  * `openhands exec "do a flip"`
* Allows users to resume a local conversation into the Terminal UI
  * `openhands resume <GUID>`
* Resume last conversation
  * `openhands resume --last`
* List available conversations to resume
  * `openhands resume`
* Ability to change output format as an argument (json, text \[default\])
  * `openhands  exec "do a flip" --output json`
* Ability to configure LLM auto-approve
  * #107 

**Nice to have**
* Ability to call and use microagents from prompts
* Ability to delegate a task to OpenHands Cloud (depends on https://github.com/All-Hands-AI/OpenHands-Cloud/issues/205)
  * `openhands -cloud -p 'do a flip'`

enyst:
Two quick notes:
- headless is heavily used for benchmark evaluations; all have instructions how to do it
- there have been some questions/discussions in slack around conceptualizing the relationship between CLI and headless the other way around: give the CLI the ability to be run non-interactively.

jpelletier1:
@enyst fair point, thank you for the feedback. I will rename this ticket to be 'Support non-interactive CLI mode' to be more clear about the intent.

[...]

PR #230, opened 2025-12-17
feat: Delegate conversation to cloud

malhotra5:
## 🚀 Feature: Cloud Conversation Support

In this PR you can run `openhands cloud -t "run a code review of this PR"` to kick off conversations in OpenHands Cloud.

We do the following

1. detect the current git repository 
2. detect the current git branch

Addresses https://github.com/All-Hands-AI/OpenHands-Cloud/issues/205 and #147

### QA
image

---

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@feat/add-cloud-conversation-support
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [ "hatchling>=1.25" ]
 
 [project]
 name = "openhands"
-version = "1.5.0"
+version = "1.6.0"
 description = "OpenHands CLI - Terminal User Interface for OpenHands AI Agent"
 readme = "README.md"
 license = { text = "MIT" }
--- a/uv.lock
+++ b/uv.lock
@@ -1466,7 +1466,7 @@ wheels = [
 
 [[package]]
 name = "openhands"
-version = "1.5.0"
+version = "1.6.0"
 source = { editable = "." }
 dependencies = [
     { name = "agent-client-protocol" },


openhands_cli/simple_main.py
```diff
@@ -142,6 +142,16 @@ def main() -> None:
             from openhands_cli.mcp.mcp_commands import handle_mcp_command
 
             handle_mcp_command(args)
+        elif args.command == "cloud":
```
enyst, line 145:
```suggestion
        elif args.command == "teleport":
```
How do you feel about “teleport”? 

“jump”
“transfer” (nope…)
“beam” (heh…)

Teleport is in CC, and it’s still new; and, it’s fun. You don’t forget it if you use it once, it will be _cool_… 🤔
Also, one could expect some users to ask; “does OH have teleport?”

malhotra5: I'm good with this! cc @jpelletier1 for visibility and if you have any opinions

enyst: Wait, I take it back… Teleport teleports the current conversation, it’s not when starting a new one…

enyst: Or it could be both? Does it make sense to use both with the meaning “restore this conversation on the cloud” and “start this conversation on the cloud”?

Seems so to me 🤔

enyst (approved):

openhands_cli/simple_main.py
malhotra5: I think we're only starting conversations on cloud for now, so I'll merge this as is

But in the future as the feature set expands, and if we restore conversations on cloud teleport might be more appropriate - wdyt?

enyst: SGTM!

PR #231, opened 2025-12-17
feat: Add tabbed interface to settings modal

malhotra5:
## Summary

This PR 

1. adds a new tab in settings page for app configuration
2. adds a new app configuration for displaying verbose outputs for token cost
3. adds callback to reload app configurations as soon as its saved

https://github.com/user-attachments/assets/c6b22cbb-3374-4270-8712-76f46f2f0635

---

---

---

---

---

---

---

---

---

---

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@feat/settings-modal-tabs
```

### DIFF
--- a/openhands_cli/refactor/modals/settings/settings_screen.py
+++ b/openhands_cli/refactor/modals/settings/settings_screen.py
@@ -100,7 +100,7 @@ class SettingsScreen(ModalScreen):
             # Tabbed content
             with TabbedContent(id="settings_tabs"):
                 # Settings Tab
-                with TabPane("Settings", id="settings_tab"):
+                with TabPane("Agent Settings", id="settings_tab"):
                     yield SettingsTab()
 
                 # CLI Settings Tab - only show if not first-time setup


openhands_cli/refactor/modals/settings/app_config.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class AppConfigSaveResult(BaseModel):
+    """Result of attempting to save app configuration."""
+
+    success: bool
+    error_message: str | None = None
+
+
+def save_app_config(config: AppConfiguration) -> AppConfigSaveResult:
```
malhotra5: @openhands please remove `save_app_config` function and `AppConfigSaveResult` class

just use .save from appconfiguration directly

openhands_cli/refactor/widgets/richlog_visualizer.py
```diff
@@ -361,6 +362,11 @@ def _create_event_collapsible(self, event: Event) -> NonClickableCollapsible | N
 
     def _format_metrics_subtitle(self) -> str | None:
         """Format LLM metrics as a visually appealing subtitle string."""
+        # Check app configuration to see if metrics should be displayed
+        app_config = AppConfiguration.load()
+        if not app_config.display_cost_per_action:
+            return None
+
```
malhotra5: @openhands note that this seems very inefficient, its reading from the filestore every single time

please make sure that we're caching or doing a more efficient update of the configuration such that we aren't constantly loading the configurations

openhands-cli.spec
```diff
@@ -61,6 +61,10 @@ a = Analysis(
         'openhands.tools.terminal',
         'openhands.tools.str_replace_editor',
         'openhands.tools.task_tracker',
+        # Include textual internal modules that might not be auto-detected
+        'textual.widgets._tab_pane',
```
enyst, line 65: This is a bit strange, was this intentional?

malhotra5: Yep! Pyinstaller tries to detect what needs to be imported and only includes those files (hence why data files like jinja templates are so likely to be left out)

Without this the e2e tests for binary executable fail (starts with a missing import error) - not sure why pyinstaller failed to include this so i manually included it here

openhands_cli/refactor/modals/settings/cli_settings.py
```diff
@@ -0,0 +1,53 @@
...
+    display_cost_per_action: bool = False
+
+    @classmethod
+    def get_config_path(cls) -> Path:
+        """Get the path to the CLI configuration file."""
+        # Use environment variable if set, otherwise use default
+        persistence_dir = os.environ.get(
+            "PERSISTENCE_DIR", os.path.expanduser("~/.openhands")
+        )
+        return Path(persistence_dir) / "cli_config.json"
```
enyst, line 22: Just want to note here, that it would be nice if we were pretty sure that we want to use both `settings` and `config` as if they were different things. If we’re not sure, maybe this one is better as `cli_settings.json`… so that we don’t have compatibility to handle.

enyst (approved):
Thank you! Interesting how the new UI develops.

I was a bit surprised by config vs settings. I think… it might be a good idea actually, if we can stick to it strictly:
- settings for agent or things from the SDK, common to all apps built on it
- configuration = application specific things.

Was that how you thought of it, or WDYT?

I will note that I had a hesitation, because for most (?) users, I’m… not sure they need or do make a distinction, so I wonder if it appears as confusing… 🤔

malhotra5:
Yep that was the idea (settings = agent or sdk settings, config = application specific things)

However, I think if this distinction isn't obvious then it will be hard to maintain. So for the sake of reducing complexity (because I would love to get this in the hands of more contributors), I think renaming to settings would be better

Will rename it!

enyst:
I actually came back to say the same, that maybe `settings` makes more sense, because apart from agent or llm, which maybe can be mapped mentally in some way, we also have in SDK some env vars, which could be exposed by the CLI to users… and there, it’s not so clear… not at all🤔

Thank you!

openhands_cli/refactor/modals/settings/cli_settings.py
malhotra5: @openhands please change `app_config` to `cli_settings` 

Do this for the files names, the class names, file paths, etc - make sure update any tests as well

run `make lint` and ensure there aren't any errors

Issue #221, opened 2025-12-17
Chore: remove old UI in favor of experimental one


malhotra5:
The tentative plan is to remove the code for the old UI in the codebase in favor the new experimental one. We have requested for feedback in [slack](https://openhands-ai.slack.com/archives/C086ARSNMGA/p1765835370551659) while the UI is still experimental

malhotra5:
The experimental UI is now the default experience - this issue is being kept open still as there are sub issues we received from feedback that requires fixing

malhotra5:
This is complete

PR #235, opened 2025-12-18
Make experimental UI the default UI

malhotra5:
## Summary

This PR

1. make the default launched UI the experimental UI
2. removes the code files for the old UI

In a follow up PR will re-organize the files to be a bit better (e.g rename the `refactor` subdirectory)

#221

---

---

---

---

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@make-experimental-ui-default
```

### DIFF
--- a/.github/workflows/update-install-website.yml
+++ b/.github/workflows/update-install-website.yml
@@ -1,0 +1,135 @@
+---
+# Workflow that updates the install-openhands-website repository when a release is published
+name: Update Install Website Version
+
+# Trigger when a release is published (not draft) or manually
+on:
+    release:
+        types: [published]
+    workflow_dispatch:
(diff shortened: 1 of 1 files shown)


enyst:
The LLMs are there! They actually got there. It made me wonder what’s the future of delegation implemented in a single SDK, when the agent can just spawn a CLI, you know… 😅

malhotra5:
OHHH that makes a ton of sense! I'm thinking `json-mode` makes a ton of sense I can get that ready pretty quickly

EDIT: i think json mode should be the default, a flag can be used to make it more silenced or perhaps verbosity can be more configurable even

PR #248, opened 2025-12-19
Chore: reorganize files

malhotra5:
This PR is a followup to re-organize files after #235

We remove the `tui` subdirectory and make it `store` instead. The `AgentStore` and `CLISettings` now belong in this folder

---

---

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@reorganize-files
```

### DIFF
--- a/tests/settings/test_skills_loading.py
+++ b/tests/settings/test_skills_loading.py
@@ -6,8 +6,6 @@ from unittest.mock import patch
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


malhotra5:
@openhands please the failing unit tests; not that this PR is reorganizing files so the tests are failing only because import paths have changed

please only modify the existing tests to fix this

PR #250, opened 2025-12-19
Fix none type error in cloud connection 

malhotra5:
---

---

---

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@fix-nonetype-error
```

### DIFF
--- a/tests/auth/test_api_client.py
+++ b/tests/auth/test_api_client.py
@@ -10,6 +10,7 @@ from openhands_cli.auth.api_client import (
     ApiClientError,
     OpenHandsApiClient,
     UnauthenticatedError,
+    _ask_user_consent_for_overwrite,
     create_and_save_agent_configuration,
     fetch_user_data_after_oauth,
 )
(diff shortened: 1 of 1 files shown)


openhands_cli/auth/api_client.py
```diff
@@ -165,6 +168,7 @@ def _ask_user_consent_for_overwrite(
         f"[/{OPENHANDS_THEME.accent}]"
     )
 
+    print("prompting user")
```
malhotra5:
```suggestion
```

malhotra5:
@openhands please read the diff on this PR and add exactly one unit test to verify that this print statement doesn't occur unless base_url exists

openhands_cli/auth/api_client.py
```diff
@@ -110,7 +111,7 @@ def _print_settings_summary(settings: dict[str, Any]) -> None:
 
 
 def _ask_user_consent_for_overwrite(
-    existing_agent,
+    existing_agent: Agent,
     new_settings: dict[str, Any],
     base_url: str = "https://llm-proxy.app.all-hands.dev/",
     default_model: str = "claude-sonnet-4-5-20250929",
```
enyst, line 117: Out of curiosity, why use `base_url` default, when we could use `openhands/` prefix? I think

malhotra5: I think sdk upstream used to automatically set the `base_url` based on the `openhands/` prefix

It no longer does that, so we have to hard code it https://github.com/OpenHands/OpenHands-CLI/pull/168

PR #254, opened 2025-12-19
feat: implement streaming LLM outputs for ACP implementation

### LINKED ISSUE #253: Feat: support streaming llm outputs
The agent sdk supports streaming llm tokens. We should consider streaming capability for CLI (the UI + ACP). CLI seems slow otherwise (even though the overall response is quick)

malhotra5:
## Summary

This PR implements token streaming for ACP only when using completion api (not responses api as it doesn't support streaming yet as per litellm)

For streamed tokens, we stream message contents, reasoning content, and tool calls. We use `streamingjson.lexer` for tool calls which have incomplete json. The `ThinkTool` is a special case where it should map to an ACP thought event rather than a tool call. 

#### Before 

https://github.com/user-attachments/assets/dcd950c4-f476-4882-abbb-81ba52b947f1

#### After

https://github.com/user-attachments/assets/08fcca09-8dc6-491b-a2a3-1cba58d6fb43

Closes #253

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@openhands/implement-streaming-llm-outputs
```

### DIFF
--- a/openhands_cli/acp_impl/agent.py
+++ b/openhands_cli/acp_impl/agent.py
@@ -97,6 +97,7 @@ class OpenHandsACPAgent(ACPAgent):
         conn: Client,
         initial_confirmation_mode: ConfirmationMode,
         resume_conversation_id: str | None = None,
+        streaming_enabled: bool = False,
     ):
         """Initialize the OpenHands ACP agent.
 
@@ -105,6 +106,7 @@ class OpenHandsACPAgent(ACPAgent):
             initial_confirmation_mode: Default confirmation mode for new sessions
             resume_conversation_id: Optional conversation ID to resume when a new
                 session is created (used with --resume flag)
+            streaming_enabled: Whether to enable token streaming for LLM outputs
         """
         self._conn = conn
         # Cache of active conversations to preserve state (pause, confirmation, etc.)
@@ -116,9 +118,11 @@ class OpenHandsACPAgent(ACPAgent):
         self._initial_confirmation_mode: ConfirmationMode = initial_confirmation_mode
         # Conversation ID to resume (from --resume flag)
         self._resume_conversation_id: str | None = resume_conversation_id
+        # Whether token streaming is enabled (from --streaming flag)
+        self._streaming_enabled: bool = streaming_enabled
         logger.info(
-            f"OpenHands ACP Agent initialized with "
-            f"confirmation mode: {initial_confirmation_mode}"
+            f"OpenHands ACP Agent initialized with confirmation mode: "
+            f"{initial_confirmation_mode}, streaming: {streaming_enabled}"
         )
         if resume_conversation_id:
             logger.info(f"Will resume conversation: {resume_conversation_id}")
@@ -254,14 +258,18 @@ class OpenHandsACPAgent(ACPAgent):
             RequestError: If MCP configuration is invalid
         """
         # Load agent specs (same as setup_conversation)
-        streaming_enabled = False
         try:
             agent = load_agent_specs(
                 conversation_id=session_id,
                 mcp_servers=mcp_servers,
                 skills=[RESOURCE_SKILL],
             )
-            streaming_enabled = not agent.llm.uses_responses_api()
+            # Streaming is enabled only if:
+            # 1. The --streaming flag was passed (self._streaming_enabled)
+            # 2. The LLM doesn't use responses API (which doesn't support streaming)
+            streaming_enabled = (
+                self._streaming_enabled and not agent.llm.uses_responses_api()
+            )
 
             if streaming_enabled:
                 # Enable streaming for llm
@@ -329,7 +337,7 @@ class OpenHandsACPAgent(ACPAgent):
             callbacks=[sync_callback],
             token_callbacks=[token_subscriber.on_token]
             if streaming_enabled
-            else None,  # Enable token streaming for completions api
+            else None,  # Enable token streaming only when --streaming flag is used
             visualizer=None,  # No visualizer needed for ACP
         )
 
@@ -756,6 +764,7 @@ class OpenHandsACPAgent(ACPAgent):
 async def run_acp_server(
     initial_confirmation_mode: ConfirmationMode = "always-ask",
     resume_conversation_id: str | None = None,
+    streaming_enabled: bool = False,
 ) -> None:
     """Run the OpenHands ACP server.
 
@@ -763,10 +772,11 @@ async def run_acp_server(
         initial_confirmation_mode: Default confirmation mode for new sessions
         resume_conversation_id: Optional conversation ID to resume when a new
             session is created
+        streaming_enabled: Whether to enable token streaming for LLM outputs
     """
     logger.info(
         f"Starting OpenHands ACP server with confirmation mode: "
-        f"{initial_confirmation_mode}..."
+        f"{initial_confirmation_mode}, streaming: {streaming_enabled}..."
     )
     if resume_conversation_id:
         logger.info(f"Will resume conversation: {resume_conversation_id}")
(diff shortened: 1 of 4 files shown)


xingyaoww:
@OpenHands /codereview-roasted

malhotra5:
@openhands please fix the critical issues outlined in the code review

enyst:
@openhands Do a /codereview-roasted on this PR. Post your findings as a comment.

enyst:
EDIT: scratching out comments addresses in followup commits


🧯 /codereview-roasted for PR #254: feat: implement streaming LLM outputs for ACP implementation

Taste Rating: 🟡 Acceptable — solid direction, but a few design warts and footguns that need tightening before this is “good taste.”

[CRITICAL ISSUES]
- ~[events/token_streamer.py, lines ~137–141] Error handling in on_token: Raising RequestError from a token callback is a great way to nuke the stream. Token callbacks should be best-effort and non-throwing. Log and surface an ACP error update instead of raising. Otherwise one malformed chunk bricks the whole run.~
- ~[events/token_streamer.py, lifecycle] Tool-call state leak: _streaming_tool_calls is never cleared. After ObservationEvent completes for a tool_call_id, the per-index ToolCallState remains resident. Over a long session, this accumulates. Fix: on ObservationEvent (completed/failed), remove any matching state(s) by tool_call_id. You currently key by index only; either track a second map by tool_call_id or scan the dict and prune.~
- ~[events/token_streamer.py, lines ~147–156] Missing field_meta on streaming updates: All stream-driven session_update calls skip metrics metadata. Non-streamed paths include it via get_metadata. UI/clients will show inconsistent status. Add field_meta=get_metadata(self.conversation) to scheduled updates.~
- ~[acp_impl/agent.py, lines ~517–523] Full traceback in RequestError payload: You switched from str(e) to full traceback. That leaks internal paths and invites client-side log spam. Keep detailed traces in logs; return a succinct reason/details to clients.~

[IMPROVEMENT OPPORTUNITIES]
- ~[events/token_streamer.py, tool call start/update] Title/kind consistency: You compute title/kind from partial args for streaming starts but from full ActionEvent in non-streaming paths. That can cause title churn and kind mismatches (e.g., file_editor view vs edit before command is fully parsed). Consider deferring start_tool_call until you have the command key, or allow one stable update when command becomes known (suppress spurious updates).~
- ~[events/token_streamer.py, multi-choice handling] You merge all choices blindly. If a provider emits multiple choices or out-of-order indices, you could interleave thought/content from different choices. At minimum, gate on choice.index==0 or group by index consistently (you already group tool_calls by index; do the same for content/reasoning, or document why choice>0 is impossible here).~
- [events/token_streamer.py, think streaming] extract_thought_piece assumes monotonic growth. Some providers backspace/modify tokens (especially “reasoning” channels). Your diffing will emit garbled deltas if the model edits earlier characters. Consider handling non-monotonic updates (e.g., LCS diff or reset on mismatch) or document/validate provider behavior.
- ~[events/token_streamer.py, start flood] You start non-think tool call immediately when first args token arrives. For noisy models that emit a tool name then hesitate, this can flicker transient tool calls that later get abandoned. Consider delaying start until you’ve seen a minimally valid args skeleton (e.g., have a command/path for terminal/file_editor) or a small debounce window.~
- ~[events/event.py + shared_event_handler.py] Two codepaths for thoughts/messages: EventSubscriber emits formatted “Reasoning/Thought” blocks; token streamer emits raw text chunks. The UX becomes inconsistent mid-run when streaming is on. Unify the formatting policy (e.g., prepend headings once, plain deltas afterward) so the transcript looks intentional, not stitched.~
- ~[tests/acp/events/test_streaming.py] Duplicate call: The test invokes subscriber.unstreamed_event_handler(obs_event) twice in a row without asserting an intermediate effect. Looks like a copy/paste slip. Clean it up or assert idempotency if that’s intended.~

[STYLE NOTES]
- ~[events/utils.py] Mixed import styles for streamingjson (module vs from Lexer). Pick one for clarity; you already wrap it well.~
- [events/token_streamer.py] The state machine is spread across three files (token_streamer, tool_state, utils) with implicit contracts. A small diagram or docstring at TokenBasedEventSubscriber explaining the lifecycle (tool_call start -> updates -> completion via ObservationEvent) would save future readers.

[BREAKING CHANGE RISK]
- Gating streaming on uses_responses_api(): Reasonable, but ensure all LLM adapters implement this. If not, you’ll accidentally disable streaming or misconfigure the LLM. Add a defensive default and a log line so operators know why streaming is on/off.

EDIT: relying on litellm's pre-defined contracts here; not addresses 

[SECURITY/CORRECTNESS]
- ~Don’t send tracebacks to clients (see Critical). Also sanitize any raw args you mirror back; terminal commands are fine but avoid echoing sensitive env if providers hallucinate them into args.~

VERDICT:
❌ Needs rework before merge. The architecture is promising, but the error-handling contract for token callbacks, cleanup of tool-call state, and metadata consistency need fixes to avoid regressions and leaks.

KEY INSIGHT:
Streaming code must be “can’t fail, won’t leak.” Make callbacks non-throwing, make state self-pruning, and keep client-visible output consistent regardless of streaming mode.

pyproject.toml
```diff
@@ -27,6 +27,8 @@ dependencies = [
   "pyperclip>=1.9.0",
   "httpx>=0.25.0",
   "textual-serve>=1.1.3",
+  "textual-dev>=1.8.0",
```
malhotra5: @openhands please move `textual-dev` dependency in the dev dependency group

rebuild the UV lock, run `make lint` and make sure its passing, pull any changes first before pushing your changes to the existing branch

malhotra5:
@enyst @xingyaoww I've addressed the feedback comments. Decided to pass over 

> [events/token_streamer.py, think streaming] extract_thought_piece assumes monotonic growth. Some providers backspace/modify tokens (especially “reasoning” channels). Your diffing will emit garbled deltas if the model edits earlier characters. Consider handling non-monotonic updates (e.g., LCS diff or reset on mismatch) or document/validate provider behavior.


I think monotonic growth is reasonable, I don't think modifying the prefix for LLMs is a thing?

> Gating streaming on uses_responses_api(): Reasonable, but ensure all LLM adapters implement this. If not, you’ll accidentally disable streaming or misconfigure the LLM. Add a defensive default and a log line so operators know why streaming is on/off.

I'm relying on LiteLLM's library for correctly identifying responses API 

lmk what you think!

openhands_cli/acp_impl/events/shared_event_handler.py
```diff
@@ -0,0 +1,205 @@
...
+            tool_call_id=event.tool_call_id,
+            status="completed",
+            text=_event_visualize_to_plain(event),
+            raw_output=event.model_dump(),
+        )
+
+    async def handle_action_event(self, ctx: _ACPContext, event: ActionEvent):
+        content = None
+        tool_kind = get_tool_kind(
+            tool_name=event.tool_name, action=getattr(event, "action", None)
```
xingyaoww: Ideally we should remove all getattr from this PR - in this case it should just be event.action

xingyaoww (approved):
Overall LGTM! I didn't look too much into details of the refactor and assumed you mostly did copy-paste the old function -- happy to get this in as long as they works

xingyaoww:
Actually i think it might be safer to cut a release without this, and we should start using this branch for a while and make sure everything is working expected before we merge it, wdyt?

malhotra5:
@xingyaoww i've cut a release, I'm thinking we merge this to main with a feature flag? Ideally we get this out sooner than later though, since we believe this will help increase the number of conversations with openhands in Toad by making the UX more snappy

xingyaoww:
@malhotra5 sounds good!

malhotra5:
@openhands please read the diff on this PR carefully and add a feature flag here for acp streaming

The streaming mode should be enabled when `openhands acp --streaming` arg is used

enyst:
> I think monotonic growth is reasonable, I don't think modifying the prefix for LLMs is a thing?
> I'm relying on LiteLLM's library for correctly identifying responses API

Sounds good btw, yes and yes. Not sure what the agent was on about - overly-defensiveness 🫠

Thank you for this! ❤️

PR #276, opened 2025-12-31
fix(input): implement auto-growing input with soft wrapping

### LINKED ISSUE #222: Experimental CLI doesn't wrap text into new lines for user input
When the user is in a single line input, once the text grows long enough it doesn't wrap. 

Proposed solution - 

1. detect when the text has grown long enough
2. automatically toggle to multi-line mode 

Cons: once toggled in multi-line mode the keybindings to submit the input, etc change

malhotra5:
## Summary

This PR ensures that when the text wraps and the input box expands vertically from top when the text grows too long in single line mode

This video demonstrates the problem

https://github.com/user-attachments/assets/58390f16-94e3-4e38-b4b0-6e378f7f6ff6

This PR is a refactor of the input widget. It simplifies the code by using a custom `TextArea` for single line inputs, where the enter key submits content or interacts with an autocomplete dropdown instead of inserting a new line.  

### Changes QAed

Features confirmed that should still work
- [X] slash commands autocomplete and selection
- [X] files and directory autocomplete and selection
- [X] toggling between single line and multi line input and preserving text
- [X] Automatically toggle to multi line input when multi line text is pasted from clipboard
- [X] ability to submit text when pressing enter in single line mode

#### Wrapping on single line input and toggling between single/multi line input while preserving text
https://github.com/user-attachments/assets/52de0d0e-81c2-428e-b604-5d378c195e01

#### Slash command and files/directory autocomplete 
https://github.com/user-attachments/assets/23a840e9-1fb6-42e8-ae4f-1b6a1e6414fe

#### Auto-toggle to multiline for multiline pastes 

https://github.com/user-attachments/assets/218cacee-dc6c-4c92-8737-92b1e2347966

Fixes #222
Fix https://github.com/OpenHands/OpenHands-CLI/issues/358

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@feat/auto-grow-input-field
```

### DIFF
--- a/tests/tui/widgets/test_richlog_visualizer.py
+++ b/tests/tui/widgets/test_richlog_visualizer.py
@@ -337,11 +337,9 @@ class TestVisualizerIntegration:
         assert collapsible.title is not None
 
         # Verify that the visualizer successfully created a collapsible widget
-        # The content should be present in some form, even if not escaped
         title_str = str(collapsible.title)
-        # Title should include a visible snippet of the XML-like command
-        assert "<function=execute_bash" in title_str
-        assert "execute_bash" in title_str  # The function name should be present
+        # Title should show the tool name (new implementation uses clean titles)
+        assert "execute_bash" in title_str
         assert len(title_str) > 0  # Title should not be empty
 
 


malhotra5:
@openhands Do a /codereview-roasted on this PR. Post your findings as a comment.

openhands_cli/refactor/widgets/user_input/expandable_text_area.py
```diff
@@ -0,0 +1,132 @@
...
+            # Multi-line content detected - notify parent
+            self.post_message(self.PasteDetected(event.text))
+            event.prevent_default()
+            event.stop()
+        # For single-line content, let the default paste behavior handle it
+
+    @property
+    def value(self) -> str:
+        """Compatibility property to match Input widget API."""
+        return self.text
```
malhotra5: @openhands It seems like we have unnecessary properties defined here

can we make sure to use `.text` directly instead of wrapping it this way? find wherever this component and make sure were using the defined attrib directly rather than redefining new setters and properties

malhotra5:
@openhands please merge main, fix any conflicts, and failing unit tests

malhotra5:
## 🔴 **Needs Improvement** - Violates Fundamental Principles

*"Bad programmers worry about the code. Good programmers worry about data structures and their relationships."*

---

### Taste Rating: 🔴 Needs Improvement

The core idea is sound - using soft wrapping with auto-grow instead of auto-toggling to multiline mode is the pragmatic choice. But the execution? Let me be direct.

---

**[expandable_text_area.py, Lines 51-52]** 🗑️ **Dead Code**: 
```python
self._cached_line_offsets: list[int] | None = None
self._cached_text_for_offsets: str | None = None
```
These fields are initialized but **never used anywhere**. This is the kind of cruft that makes codebases rot. Either use them or delete them. Don't leave corpses lying around.

**[input_field.py, Lines 23-254]** 🏗️ **God Class Alert**:
This class is doing the work of at least 3 different components:
1. Managing two different text areas (single-line and multiline)
2. Handling autocomplete coordination
3. Mode toggling logic
4. Paste detection and handling
5. Submission logic

This violates single responsibility so hard it hurts. When a class needs 250+ lines and handles this many concerns, it's a design smell. The fact that you need to track `is_multiline_mode` as state and constantly check it everywhere is a red flag that the abstraction is wrong.

---

**[input_field.py, Line 101]** 📝 **Typo in Public API**:
```python
self.mutliline_mode_status = Signal(self, "mutliline_mode_status")
```
"mutliline" → "multiline". This is a Signal that external code might subscribe to. Typos in public APIs are embarrassing and painful to fix later.

**[input_field.py, Lines 242-253]** 🤔 **Suspicious Paste Handling**:
```python
def on_paste_detected(self, event: AutoGrowTextArea.PasteDetected) -> None:
    if not self.is_multiline_mode:
        self.input_widget.insert(
            event.text,
            self.input_widget.cursor_location,
        )
        self.action_toggle_input_mode()
```
You insert the text into the single-line widget, then immediately toggle to multiline mode. What happens to the text you just inserted? Does it get preserved correctly? This feels like it could have edge cases. The comment says "Get current text and cursor position before switching modes" but you're not actually doing that - you're inserting first, then switching.

**[text_area_with_autocomplete.py, Lines 15-30]** 🔧 **Orphaned Function**:
`detect_completion_type()` is a module-level function that's only used by `TextAreaAutoComplete.update_candidates()`. It should be a method on the class, not floating around at module level. This isn't a library - keep related logic together.

**[expandable_text_area.py, Lines 64-73]** ⚠️ **Special Case Handling**:
```python
async def _on_key(self, event) -> None:
    if event.key == "enter":
        self.post_message(self.EnterPressed())
        event.prevent_default()
        event.stop()
        return
    await super()._on_key(event)
```
Intercepting Enter to post a message is a special case. The parent `InputField` then has to handle this message and decide what to do. This is the kind of indirection that makes code hard to follow. Consider whether the TextArea should just handle submission directly, or whether there's a cleaner composition pattern.

---

**[models.py]** 🤷 **Pydantic for Simple Data**:
Using Pydantic with `frozen=True` for what is essentially a named tuple is overkill. A `@dataclass(frozen=True)` or even a `NamedTuple` would be simpler and have no runtime dependency. But this is a style preference, not a bug.

**[textual_app.tcss, Lines 41-47]** 📐 **Magic Numbers**:
```css
#input_area {
    height: auto;
    max-height: 15;
    ...
}
```
Why 15? Why not 14 or 16? Magic numbers without explanation make future maintainers guess. A comment would help.

---

### **VERDICT:** ❌ **Needs Rework**

The dead code and the God class are the main issues. The approach is correct - soft wrapping with auto-grow is the right solution to the problem. But the implementation has accumulated complexity that will make this hard to maintain.

---

### **KEY INSIGHT:**

*The `InputField` class is trying to be two different widgets (single-line and multiline) at once, and the complexity of managing that duality has leaked into every method. Consider whether you actually need both modes, or if the auto-growing TextArea with soft wrap is sufficient on its own.*

---

*"Talk is cheap. Show me the code."* - And when you show me dead code, I'll tell you to delete it. 🔥

malhotra5:
@openhands /codereview-roasted. post the review back as a github comment

malhotra5:
## 🔥 Code Review - Linus Style

### Taste Rating: 🟡 **Acceptable** - Works but could be cleaner

Alright, let's talk about this PR. The overall architecture is reasonable - you've separated concerns into models, input widgets, and autocomplete. That's good taste. But there are some things that make me twitch.

---

**[single_line_input.py, Line 8] 🔤 Spelling**: `SingleLineInputWithWraping` - Did we forget how to spell "Wrapping"? This is a public class name that will haunt you forever. Same goes for `MutliLinePasteDetected` on Line 18 - it's "Multi", not "Mutli". These typos will propagate through your entire codebase and make everyone who reads it question your attention to detail.

```python
# Current (wrong)
class SingleLineInputWithWraping(TextArea):
class MutliLinePasteDetected(Message):

# Should be
class SingleLineInputWithWrapping(TextArea):
class MultiLinePasteDetected(Message):
```

**[input_field.py, Lines 53-54, 67-68, 71-72] 💥 CSS `!important` Abuse**: You're using `!important` on borders like it's going out of style. This is a code smell that screams "I don't understand CSS specificity and I'm just hammering until it works." If you need `!important` in your own component's DEFAULT_CSS, your CSS architecture is broken.

```css
/* Current (fighting with yourself) */
border: solid $primary !important;

/* Should be: Fix the specificity properly or understand why you're fighting */
border: solid $primary;
```

---

**[autocomplete_dropdown.py, Line 272] 🎯 Magic Tuple**: `move_cursor((0, len(self.single_line_widget.text)))` - What is `(0, len(text))`? Row 0, column N? This assumes single-line behavior but isn't self-documenting. A comment or named constant would help future-you understand this.

**[autocomplete_dropdown.py, Lines 142-158] 🔀 elif Chain**: The `process_key` method uses a long if/elif chain. This is acceptable for now, but if you add more keys, consider a dispatch dictionary:

```python
# Current (acceptable but not scalable)
if key == "down":
    ...
elif key == "up":
    ...

# Better for growth
key_handlers = {
    "down": self._handle_down,
    "up": self._handle_up,
    ...
}
```

**[input_field.py, Line 96] 📛 Signal Naming**: `multiline_mode_status` is a noun, but signals typically represent events. Consider `multiline_mode_changed` to indicate something happened.

**[single_line_input.py, Lines 61-70] ⚡ Async Key Handler**: The `_on_key` method is async but you're calling `event.prevent_default()` and `event.stop()` synchronously before returning. Make sure this plays nicely with Textual's event loop - async handlers that return early can have subtle timing issues.

---

### **[STYLE NOTES]** (Minor - only mention if genuinely important)

**[models.py] ✅ Good**: Using Pydantic with `frozen=True` for immutable completion items is the right call. Clean data modeling.

**[textual_app.tcss, Lines 41-48] ✅ Good**: The `height: auto` with `max-height: 15` and `align: left bottom` is the correct approach for auto-growing input that expands upward.

**[autocomplete_dropdown.py, Line 152] 🧹 Dead Code**: There's a commented-out line `# self.post_message(self.CompletionSelected(item))`. Either implement it or remove it. Commented code is technical debt.

---

### **VERDICT:**
✅ **Worth merging** (after fixing the typos): Core logic is sound, the architecture is reasonable, and it solves the real problem of text wrapping in single-line mode. The typos in class names are embarrassing but fixable. The `!important` abuse should be addressed but isn't blocking.

### **KEY INSIGHT:**
The fundamental approach of using a TextArea with `soft_wrap=True` and `height: auto` instead of toggling to multiline mode is the right solution - it's simpler and maintains consistent keybindings. Just clean up the spelling before this becomes permanent API.

---

*"Talk is cheap. Show me the code." - And please spell-check your class names.*

malhotra5:
I've addressed the comments from the code review roasted

malhotra5:
@openhands please fix the merge conflicts

PR #280, opened 2026-01-02
fix: use model_id without provider prefix when loading settings

neubig:
## Summary

Fixes `InvalidSelectValueError: Illegal select value 'cerebras/zai-glm-4.6'` that occurs when opening the settings screen after configuring a model like `cerebras/zai-glm-4.6`.

## Root Cause

When loading settings in basic mode, the `_load_current_settings()` method was setting `model_select.value` to the full model string (e.g., `cerebras/zai-glm-4.6`), but the dropdown options from `get_model_options()` only contain model IDs without the provider prefix (e.g., `zai-glm-4.6`).

This caused the Textual Select widget to throw `InvalidSelectValueError` because the value being set was not in the list of valid options.

## Fix

Changed line 177 in `settings_screen.py` to use the `model` variable (extracted from splitting `llm.model` by `/`) instead of the full `llm.model` string when setting `model_select.value`.

## Testing

HUMAN: this has been QAed, confirmed to fail before

Screenshot 2026-01-02 at 9 22 29 AM

and work after

Screenshot 2026-01-02 at 9 23 02 AM

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@fix/model-select-invalid-value
```

### DIFF
--- a/openhands_cli/tui/modals/settings/settings_screen.py
+++ b/openhands_cli/tui/modals/settings/settings_screen.py
@@ -174,7 +174,7 @@ class SettingsScreen(ModalScreen):
 
                 # Update model options and select current model
                 self._update_model_options(provider)
-                # Use model (without provider prefix) since dropdown options don't include prefix
+                # Use model without provider prefix (dropdown options don't have it)
                 self.model_select.value = model
 
         # API Key (show masked version)
--- a/tests/refactor/modals/settings/test_settings_screen.py
+++ b/tests/refactor/modals/settings/test_settings_screen.py
@@ -132,7 +132,7 @@ async def test_load_current_settings_basic_mode(
     screen.current_agent = fake_agent_store.load()
 
     with patch.object(ss, "get_model_options") as mock_get_options:
-        # Model options don't include provider prefix (matches real get_model_options behavior)
+        # Model options don't include provider prefix
         mock_get_options.return_value = [
             ("gpt-4o-mini", "gpt-4o-mini"),
             ("gpt-4o", "gpt-4o"),


neubig:
@openhands fix the failing pre-commit

neubig:
@openhands fix failing pre-commit

openhands_cli/tui/modals/settings/settings_screen.py
```diff
@@ -174,7 +174,8 @@ def _load_current_settings(self) -> None:
 
                 # Update model options and select current model
                 self._update_model_options(provider)
-                self.model_select.value = llm.model
+                # Use model without provider prefix (dropdown options don't have it)
+                self.model_select.value = model
```
enyst, line 178: Thank you! Out of curiosity, why did it work for others? The code is in "basic mode" for existing `settings`, which... should be widely in use. I'm a bit wondering if this bug isn't some `cerebras` special case, and I hope we don't break the others?

neubig: I think the new CLI interface is pretty new, so it can be attributed to that perhaps?

Also, maybe we were using models that didn't require a prefix. Not 100% sure though.

enyst: I think most of the time (all the time?) there was no need for provider string. Because the model name contains it (in the SDK).  🤔

enyst: > I think the new CLI interface is pretty new, so it can be attributed to that perhaps?

Yes... I think that's likely - how does it work for other models?

enyst: One reason I ask is openai-compatible models: litellm has this, a bit arcane, rule that it needs "openai/" in front of openai-compatible providers, to know how to make transformations of the request and to route. But if we split "openai" and consider it _the OpenAI models' provider_, then we would only assign it the actual OpenAI models, so we can't do that anymore, if that makes sense?
(sorry for writing on the run)

neubig:
I confirmed that it works with anthropic as well!

PR #284, opened 2026-01-02
feat: Add Ctrl+U shortcut to clear input field

### LINKED ISSUE #282: Add "delete all" keyboard shortcut for input box
## Feature Request

When accidentally pasting a large amount of text into the input box, it can be difficult and time-consuming to delete all of it.

## Proposed Solution

Add a "delete all" or "clear input" keyboard shortcut (e.g., `Ctrl+U` or similar) that allows users to quickly clear the entire input box content.

## Use Case

Users sometimes accidentally paste large amounts of text into the input box. Currently, deleting all this text requires holding backspace or manually selecting all text. A dedicated keyboard shortcut would significantly improve the user experience in these situations.

Boecc:
### Description
This PR addresses issue #282 by adding a keyboard shortcut (`Ctrl+U`) to quickly clear the input field.

### Changes
- Added `Ctrl+U` to the `BINDINGS` list in `InputField`.
- Implemented `action_clear_input` method to handle the clear action for both single-line (`Input`) and multi-line (`TextArea`) modes.

### Testing
**Note:** I am developing on Windows and could not run the TUI locally due to the `fcntl` dependency issue (OpenHands-CLI/issues/282 context).
However, the implementation follows standard Textual patterns. I would appreciate if a maintainer could verify the behavior on a compatible environment (Linux/Mac).

Fixes #282

### DIFF
--- a/openhands_cli/tui/widgets/input_field.py
+++ b/openhands_cli/tui/widgets/input_field.py
@@ -14,6 +14,9 @@ from openhands_cli.tui.widgets.autocomplete import EnhancedAutoComplete
 
 class PasteAwareInput(Input):
     """Custom Input widget that can handle paste events and notify parent."""
+    BINDINGS = [
+        Binding("ctrl+a", "select_all", "Select all"),
+    ]
 
     class PasteDetected(Message):
         """Message sent when multi-line paste is detected."""
@@ -32,12 +35,16 @@ class PasteAwareInput(Input):
             event.stop()
         # For single-line content, let the default paste behavior handle it
 
+class SelectAwareTextArea(TextArea):
+    """TextArea that overrides Ctrl+A to select all."""
+    BINDINGS = [
+        Binding("ctrl+a", "select_all", "Select all"),
+    ]
 
 class InputField(Container):
     BINDINGS: ClassVar = [
         Binding("ctrl+l", "toggle_input_mode", "Toggle single/multi-line input"),
         Binding("ctrl+j", "submit_textarea", "Submit multi-line input"),
-        Binding("ctrl+a", "select_all", "Select all"),
     ]
 
     DEFAULT_CSS = """
@@ -99,7 +106,7 @@ class InputField(Container):
         yield self.input_widget
 
         # Multi-line textarea (initially hidden)
-        self.textarea_widget = TextArea(
+        self.textarea_widget = SelectAwareTextArea(
             id="user_textarea",
             soft_wrap=True,
             show_line_numbers=False,
@@ -153,14 +160,6 @@ class InputField(Container):
                 self.action_toggle_input_mode()
                 # Submit the content
                 self.post_message(self.Submitted(content))
-
-    def action_select_all(self) -> None:
-        """Select all content in the active input widget."""
-        if self.is_multiline_mode:
-            self.textarea_widget.select_all()
-        else:
-            self.input_widget.action_end()
-            self.input_widget.select_all()
 
     def on_input_submitted(self, event: Input.Submitted) -> None:
         """Handle single-line input submission."""


malhotra5:
Thanks for the quick contribution! I wonder if support `ctrl + a` to select all the text would be a better interaction (its a more familiar pattern to do `ctrl +a` and then press the delete or backspace key)

Boecc:
That's a great point. Ctrl+A is definitely a more familiar pattern for most users and adds more utility. I will update the PR to implement 'Select All' functionality instead.

malhotra5:
Hmm I tried testing the functionality locally

It seems ctrl + a isn't selecting all the text, rather its just moving the cursor all the way to the left of the current line


@openhands please read the PR and investigate why

Note that you are on a PR originating from a forked repo. Don't make any changes please use the $GITHUB_TOKEN and post back your findings

malhotra5:
## Investigation Findings

I investigated why `Ctrl+A` is moving the cursor to the beginning of the line instead of selecting all text.

### Root Cause

The issue is that **Textual's `Input` widget has a built-in binding for `ctrl+a`** that maps to the `home` action (move cursor to start of line). This is a standard readline/emacs keybinding.

From Textual's Input widget default bindings:
```python
Binding(key='home,ctrl+a', action='home', description='Go to start')
Binding(key='ctrl+shift+a', action='select_all', description='Select all')
```

Similarly, the `TextArea` widget also has:
```python
Binding(key='home,ctrl+a', action='cursor_line_start', description='Cursor line start')
```

### Why the PR's approach doesn't work

The PR adds the `ctrl+a` binding to the `InputField` container:
```python
Binding("ctrl+a", "select_all", "Select all")
```

However, when the `Input` or `TextArea` widget has focus, **the focused widget's bindings take precedence** over the parent container's bindings. So the Input widget's `ctrl+a → home` binding fires before the InputField's binding can be reached.

### Possible Solutions

1. **Use `Ctrl+Shift+A`** - This is already the default Textual binding for "select all" in the Input widget, so it would work out of the box without any code changes.

2. **Override the binding in `PasteAwareInput`** - Add a custom binding in the `PasteAwareInput` class that overrides the default `ctrl+a` behavior:
   ```python
   class PasteAwareInput(Input):
       BINDINGS = [
           Binding("ctrl+a", "select_all", "Select all"),
       ]
   ```

3. **Go back to the original `Ctrl+U` approach** - The original issue #282 suggested `Ctrl+U` for clearing input. Interestingly, Textual's Input widget already has `ctrl+u` bound to `delete_left_all` (delete all to the left), which combined with first moving to the end would effectively clear the input.

### Recommendation

Given the original issue was about quickly clearing/deleting input content, I'd suggest either:
- Sticking with `Ctrl+U` (which already has partial support in Textual for deleting left)
- Using `Ctrl+Shift+A` for select all (already works by default in Textual's Input widget)

If you want `Ctrl+A` specifically, you'll need to override the binding in the `PasteAwareInput` class (and create a similar wrapper for TextArea) to prevent the default `home` action from taking precedence.

Boecc:
Thanks for the investigation! Since the goal is to provide a familiar pattern for users (who expect `Ctrl+A` to select all), Option 2 (overriding the binding in `PasteAwareInput`) seems like the best path forward to ensure excellent UX.

Should I proceed with implementing the override for `PasteAwareInput`?

malhotra5:
Hi @Boecc, yes I think overriding the default binding is the way to go!

malhotra5:
It seems ctrl + a is moving the cursor to the left of the input field. CMD + a on mac selects the entire screen instead of just the text inside the component


https://github.com/user-attachments/assets/c03d394e-c214-4709-a440-e1fad4d4b65c

Issue #275, opened 2025-12-31
Action confirmation pane doesn't have visible buttons when content is too long


lily11111222:
In the new UI of the Cli of version 1.7.0, when the `file_editor` on the right side is too long, there is no option for confirmation and it gets stuck here...

<img width="2780" height="1746" alt="Image" src="https://github.com/user-attachments/assets/f384e038-d628-4a1c-94ea-eebbe730654d" />

lily11111222:
so can you please tell me what can I do now to confirm this action step? Is there any shortcut key to do this?

malhotra5:
It seem you can't confirm the action at this point, apologies for this!

A workaround is to close the app and resume the conversation with `--llm-approve` or `--always--approve`

For example `openhands --resume <conv-id> --always-approve` and you can prompt the agent to "please continue" from there

lily11111222:
ok, I'll give it a try, thanks.

malhotra5:
@openhands please review the description of this issue and fix the problem 

The confirmation pane contents are not visible when the screen is too small and users seem to be unable to scroll to see the entire contents (buttons are hidden, contents that spill can't be scrolled to, etc)

malhotra5:
Resolved in the latest release!

PR #297, opened 2026-01-05
Fix confirmation panel scrolling when content is too long

malhotra5:
## Summary
This PR

1. renames the test dir folder `refactor` to `tui` 
2. fixes the bug where the confirmation side panel wasn't scrollable 
3. adds tests for the confirmation side panel 

Related to #275

https://github.com/user-attachments/assets/241a9bbe-4cec-49fa-8e2f-ef4036dcef37

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@openhands/fix-confirmation-panel-scrolling
```

### DIFF
--- a/tests/tui/panels/test_confirmation_panel.py
+++ b/tests/tui/panels/test_confirmation_panel.py
@@ -1,10 +1,12 @@
 """Tests for confirmation panel functionality."""
 
+from __future__ import annotations
+
 from unittest import mock
 
 import pytest
 from textual.app import App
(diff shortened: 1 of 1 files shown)


xingyaoww (approved):
LGTM, although i'd personally prefer it to not be on the right side panel, but rather stay under the action like the old way :(

malhotra5:
Gotten lots of feedback regarding this, I think we should definitely move it under the action 


Patching this quickly for now, will open an issue for this

PR #299, opened 2026-01-06
Move confirmation panel inline underneath action events

### LINKED ISSUE #298: Move the confirmation panel to underneath the action event collapsible
We have an action confirmation pane that opens to the right side of the screen. We should do the following

1. move the action confirmation underneath the latest action event that appears in the main page but above the input field
2. since the action event already includes the action details, we simply need to only render the confirmation options
3. once an option is selected, the action pane should disappear 
4. remove the side panel for the confirmation panel

malhotra5:
## Summary

This PR moves the action confirmation panel from a side panel to an inline panel that appears underneath the latest action event in the main display area.

### 1. Replace `ConfirmationSidePanel` with `InlineConfirmationPanel`
- Created a new `InlineConfirmationPanel` widget that only shows confirmation options
- The action details are already visible in the action event collapsible above, so we don't need to duplicate them

### 2. Mount inline confirmation in main_display
- The confirmation panel is now mounted in the `main_display` (VerticalScroll) area
- It appears underneath the latest action event collapsible but above the input field
- The display automatically scrolls to show the confirmation panel

### 3. Simplified confirmation options
- Shows a compact horizontal layout with the prompt and options
- Options: ✅ Yes, ❌ No, 🔄 Always, ⚠️ Auto LOW/MED

### 4. Panel disappears after selection
- Once an option is selected, the inline confirmation panel is removed from the display
- This keeps the UI clean and focused

### 5. Removed side panel approach
- Removed `ConfirmationSidePanel` class
- Updated styles for the new inline layout
- Updated all related tests

## Testing

- All 544 existing tests pass
- Updated tests for the new `InlineConfirmationPanel` widget
- Linting passes

Fixes #298

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/b0b78e915c874c91a8ee05128391aa7c)

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@openhands/move-confirmation-panel-inline
```

### DIFF
--- a/openhands_cli/tui/panels/confirmation_panel.py
+++ b/openhands_cli/tui/panels/confirmation_panel.py
@@ -1,15 +1,41 @@
 """Confirmation panel for displaying user confirmation options inline."""
 
 from collections.abc import Callable
+from typing import ClassVar
 
 from textual.app import ComposeResult
-from textual.containers import Container, Horizontal
+from textual.containers import Container, Vertical
 from textual.widgets import ListItem, ListView, Static
 
 from openhands_cli.tui.panels.confirmation_panel_style import (
     INLINE_CONFIRMATION_PANEL_STYLE,
 )
 from openhands_cli.user_actions.types import UserConfirmation
+
+
+class ConfirmationOption(Static):
+    """A confirmation option that shows > when highlighted."""
+
+    def __init__(self, label: str, **kwargs):
+        super().__init__(**kwargs)
+        self.label = label
+        self.is_highlighted = False
+
+    def on_mount(self) -> None:
+        """Set initial display."""
+        self._update_display()
+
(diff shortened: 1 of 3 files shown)


malhotra5:
Tested that this works

image

image

xingyaoww:
@malhotra5 could we get rid of the boxes inside the yellow box? So we can expect a UI similar to the UI before the UI refactor?

malhotra5:
@openhands please address the feedback 

Get rid of the boxes/borders on the confirmation. Have a single `>` mark for each option

malhotra5:
Verified we're using a similar UI to before. This will free up the right side of the page for planning related features  


image

PR #304, opened 2026-01-07
Fix AGENTS.md not being loaded when skills directories don't exist

### LINKED ISSUE #300: AGENTS.md are not loaded by the CLI
AGENTS.md are not loaded by the CLI - i.e. doesn't appear in the system prompt

xingyaoww:
## Summary

Fixes #300 - AGENTS.md files are not loaded by the CLI (i.e., don't appear in the system prompt).

## Problem

The root cause is a bug in the SDK's `load_project_skills()` function. Third-party skill files like AGENTS.md are only checked inside `load_skills_from_dir()`, but that function is only called if the `.openhands/skills` or `.openhands/microagents` directories exist.

This means if a project has an AGENTS.md file but no `.openhands/skills` directory, the AGENTS.md file is never loaded.

## Solution

This PR adds a workaround in the CLI to ensure third-party skill files are loaded regardless of whether the skills directories exist:

1. Added a new function `load_third_party_skills_from_work_dir()` that:
   - Directly scans the work directory for third-party skill files
   - Uses case-insensitive matching (handles AGENTS.md, agents.md, Agents.md, etc.)
   - Supports all known third-party files: AGENTS.md, AGENT.md, CLAUDE.md, GEMINI.md, .cursorrules

2. Integrated this function into `AgentStore.load()` to merge third-party skills with project skills, avoiding duplicates.

## Tests Added

- `test_load_agents_md_without_skills_directory`: Main bug fix test - verifies AGENTS.md is loaded even when `.openhands/skills` doesn't exist
- `test_load_agents_md_with_skills_directory`: Verifies no duplicates when both AGENTS.md and skills directory exist
- `test_load_third_party_skills_function_directly`: Unit test for the new function
- `test_load_third_party_skills_empty_directory`: Edge case test for empty directory
- `test_load_third_party_skills_case_insensitive`: Verifies case-insensitive file matching
- `test_load_multiple_third_party_files`: Tests loading multiple third-party files

## Note on SDK Bug

The underlying bug is in the SDK's `load_project_skills()` function in `openhands/sdk/context/skills/skill.py`. A proper fix should be made in the SDK to check for third-party files even when skills directories don't exist. This PR provides a workaround in the CLI until the SDK is fixed.

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/83f3dc2ce58142c0a793856525bf898f)

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@openhands/fix-agents-md-loading
```

### DIFF
--- a/openhands_cli/stores/agent_store.py
+++ b/openhands_cli/stores/agent_store.py
@@ -1,6 +1,8 @@
 # openhands_cli/settings/store.py
 from __future__ import annotations
 
+import logging
+from pathlib import Path
 from typing import Any
 
 from prompt_toolkit import HTML, print_formatted_text
@@ -13,6 +15,7 @@ from openhands.sdk import (
     LocalFileStore,
 )
 from openhands.sdk.context import load_project_skills
+from openhands.sdk.context.skills import Skill, load_skills_from_dir
 from openhands.tools.preset.default import get_default_tools
 from openhands_cli.locations import (
     AGENT_SETTINGS_PATH,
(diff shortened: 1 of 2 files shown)


xingyaoww:
Closing this PR in favor of fixing the bug in the SDK directly: https://github.com/OpenHands/software-agent-sdk/pull/1624

The SDK fix is cleaner because it modifies `load_project_skills()` to always check for third-party skill files (AGENTS.md, etc.) directly from the work directory, regardless of whether the skills directories exist. This leverages the existing `find_third_party_files()` utility which already does case-insensitive matching.

PR #322, opened 2026-01-09
fix: Correctly format LLM model names for litellm providers

pashkash:
## Summary

Fixes LLM provider routing issues when using providers like OpenRouter, Nvidia NIM, DeepInfra, etc. where model names already contain `/` (e.g., `openai/gpt-4.1` from OpenRouter).

Also cleans up the provider list by filtering out invalid "providers" that are actually vendor names.

## Problem

When selecting a provider like `openrouter` with model `google/gemini-3-flash-preview`, the model name was saved as `google/gemini-3-flash-preview` instead of the required `openrouter/google/gemini-3-flash-preview`. This caused litellm to fail with:

```
LLM Provider NOT provided. Pass in the LLM provider you are trying to call.
```

Additionally, the provider dropdown contained ~116 entries including invalid "providers" like `meta-llama`, `other`, `1024-x-1024` that were parsed from model names in the SDK.

### 1. `openhands_cli/tui/modals/settings/utils.py`
- Always add provider prefix in `get_full_model_name()` regardless of `/` in model name
- This matches litellm documentation for all providers

### 2. `openhands_cli/tui/modals/settings/choices.py`
- Always include VERIFIED_MODELS providers (`openhands`, `openai`, `anthropic`, `mistral`)
- Filter UNVERIFIED providers to only include valid litellm providers
- Sort providers and models alphabetically, remove duplicates

### 3. `tests/tui/modals/settings/test_settings_utils.py`
- Added test cases for openrouter, nvidia_nim, and deepinfra providers

## Testing

All 603 tests pass.

## Before/After

**Provider list:** 116+ entries → 60 valid providers (sorted alphabetically)

| Provider | Model | Before | After |
|----------|-------|--------|-------|
| openrouter | `google/gemini-3-flash-preview` | `google/gemini-3-flash-preview` ❌ | `openrouter/google/gemini-3-flash-preview` ✅ |
| nvidia_nim | `meta/llama3-70b-instruct` | `meta/llama3-70b-instruct` ❌ | `nvidia_nim/meta/llama3-70b-instruct` ✅ |
| openhands | `claude-sonnet-4-5-20250929` | `openhands/claude-sonnet-4-5-20250929` ✅ | `openhands/claude-sonnet-4-5-20250929` ✅ |

### DIFF
--- a/openhands_cli/tui/modals/settings/choices.py
+++ b/openhands_cli/tui/modals/settings/choices.py
@@ -1,20 +1,49 @@
+import litellm
+
 from openhands.sdk.llm import UNVERIFIED_MODELS_EXCLUDING_BEDROCK, VERIFIED_MODELS
 
 
+# Get set of valid litellm provider names for filtering
+# See: https://docs.litellm.ai/docs/providers
+_VALID_LITELLM_PROVIDERS: set[str] = {
+    str(getattr(p, "value", p)) for p in litellm.provider_list
+}
+
+
 def get_provider_options() -> list[tuple[str, str]]:
-    """Get list of available LLM providers."""
-    providers = list(VERIFIED_MODELS.keys()) + list(
-        UNVERIFIED_MODELS_EXCLUDING_BEDROCK.keys()
-    )
-    return [(provider, provider) for provider in providers]
+    """Get list of available LLM providers.
+
+    Includes:
+    - All VERIFIED_MODELS providers (openhands, openai, anthropic, mistral)
+      even if not in litellm.provider_list (e.g. 'openhands' is custom)
+    - UNVERIFIED providers that are known to litellm (filters out invalid
+      "providers" like 'meta-llama', 'Qwen' which are vendor names)
+
+    Sorted alphabetically.
+    """
+    # Verified providers always included (includes custom like 'openhands')
+    verified_providers = set(VERIFIED_MODELS.keys())
+
+    # Unverified providers are filtered to only valid litellm providers
+    unverified_providers = set(UNVERIFIED_MODELS_EXCLUDING_BEDROCK.keys())
+    valid_unverified = unverified_providers & _VALID_LITELLM_PROVIDERS
+
+    # Combine and sort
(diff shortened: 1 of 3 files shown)


openhands_cli/tui/modals/settings/utils.py
```diff
@@ -79,10 +79,11 @@ def get_full_model_name(self):
...
         model_str = str(self.model)
-        full_model = (
-            f"{self.provider}/{model_str}" if "/" not in model_str else model_str
-        )
-        return full_model
+
+        # Always add provider prefix - litellm requires it for routing.
+        # Even if model contains '/' (e.g. "openai/gpt-4.1" from openrouter)
+        # See: https://docs.litellm.ai/docs/providers
+        return f"{self.provider}/{model_str}"
```
enyst, line 86: This makes perfect sense, yes; there are multiple providers like that, which proxy to others, basically

openhands_cli/tui/modals/settings/choices.py
```diff
@@ -1,20 +1,49 @@
...
+      "providers" like 'meta-llama', 'Qwen' which are vendor names)
+
+    Sorted alphabetically.
+    """
+    # Verified providers always included (includes custom like 'openhands')
+    verified_providers = set(VERIFIED_MODELS.keys())
+
+    # Unverified providers are filtered to only valid litellm providers
+    unverified_providers = set(UNVERIFIED_MODELS_EXCLUDING_BEDROCK.keys())
+    valid_unverified = unverified_providers & _VALID_LITELLM_PROVIDERS
```
enyst, line 29: @openhands investigate where do unverified models come from, in the sdk, and make an issue on the sdk repo to clean those up / verify as best we can, e.g. maybe we should refactor this code to be there.
(clone yourself agent-sdk from the same org)

enyst: SDK PR for the root cause:
- https://github.com/OpenHands/software-agent-sdk/pull/1668

PR #341, opened 2026-01-12
Feat: delegate tool for spawning sub-agents

malhotra5:
## Summary

This PR adds the delegate tool to the agent to support sub-agents. Acceptance criteria considered is 

- [x] Can delegate sub-agent in TUI
- [x] Can delegate sub-agent in ACP
- [x] Delegate tool call is streamed in tokens
- [x] Can resume conversations that didn't previously contain the delegate tool 
- [X] Can see subagent events 

### Delegate in TUI 

https://github.com/user-attachments/assets/761cb961-140f-4221-a871-8f2890045fd2

### Delegate in ACP + streamed tool calls
https://github.com/user-attachments/assets/c381cfd1-e545-444d-a6c5-1832a4272f2a

### Can resume convo that didn't contain delegate tool
https://github.com/user-attachments/assets/d453ffcc-9e08-4d8d-ac41-ca861b7cbbec

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@add-delegate-tool-support
```

### DIFF
--- a/tests/tui/widgets/test_richlog_visualizer.py
+++ b/tests/tui/widgets/test_richlog_visualizer.py
@@ -815,7 +815,6 @@ class TestPlanPanelIntegration:
                 with patch.object(app.plan_panel, "toggle") as mock_toggle:
                     visualizer._do_refresh_plan_panel()
                     mock_toggle.assert_called_once()
-push your changes to the existing branch
 
 
 class TestSubVisualizerCreation:


malhotra5:
## 🔴 **Needs improvement** - Violates fundamental principles

*Adjusts reading glasses and sighs deeply*

Look, I appreciate the enthusiasm for adding delegate tool support, but this PR has some fundamental issues that need addressing before it's ready for prime time.

---

**[openhands_cli/*, Multiple Files] 🔨 Code Duplication - The Cardinal Sin**

You've implemented the *exact same title generation logic* in **THREE different places**:

1. `tool_state.py` (lines 153-167)
2. `utils.py` (lines 203-210)
3. `richlog_visualizer.py` (lines 225-254)

This is textbook "bad taste." When you fix a bug in one place, you'll forget the other two. When you add a new delegate command, you'll need to update three files. This is how technical debt compounds into bankruptcy.

**Fix:** Extract a single `format_delegate_title(action)` helper and use it everywhere. Data structures and their formatting should live in ONE place.

---

**[richlog_visualizer.py, Lines 225-254] 🔨 Complexity - 4+ Levels of Nesting**

```python
if isinstance(action, DelegateAction):
    if action.command == "spawn":
        for i, agent_id in enumerate(ids):
            if agent_type and agent_type != "default":
                # We're 4 levels deep here
```

"If you need more than 3 levels of indentation, you're screwed." This block is doing too much. The spawn case alone has nested conditionals inside a loop inside conditionals.

**Fix:** Extract `_format_spawn_title()` and `_format_delegate_title()` methods. Each should be ~5 lines max.

---

**[PR Description vs Actual Changes] 🔨 Misleading Documentation**

The PR description says:
> "Added import for `DelegateTool` to register the delegate tool when the CLI is loaded"

But `setup.py` actually **REMOVES** the tool imports:
```diff
-from openhands.tools.file_editor import FileEditorTool  # noqa: F401
-from openhands.tools.task_tracker import TaskTrackerTool  # noqa: F401
-from openhands.tools.terminal import TerminalTool  # noqa: F401
```

The registration now happens implicitly through `utils.py` imports. This is confusing - either update the description or make the registration explicit and documented.

---

**[richlog_visualizer.py, Line 228] 🔍 Defensive Programming**

```python
agent_types = action.agent_types or []
```

You're assuming `action.agent_types` exists. If `DelegateAction` changes upstream and this attribute is removed or renamed, you'll get an `AttributeError`. Meanwhile, in `tool_state.py` you properly check `isinstance(ids, list)`.

**Fix:** Be consistent. Either trust the type system everywhere or validate everywhere.

---

**[utils.py, Lines 203-210 vs richlog_visualizer.py, Lines 225-254] 🔍 Inconsistent Formatting**

The titles generated are subtly different:

- `utils.py`: `"Spawning 2 sub-agent(s): research, implementation"`
- `richlog_visualizer.py`: `"Spawn Sub-agents: research (researcher), implementation"`

One includes agent types, one doesn't. One says "Spawning", one says "Spawn Sub-agents". Pick ONE format and use it consistently.

---

**[tool_state.py, Lines 153-167] 🔍 Special Case Handling**

```python
if command == "spawn":
    # 8 lines of logic
elif command == "delegate":
    # 6 lines of logic
return "Delegate"
```

This pattern of `if/elif/else` for command types will grow as you add more delegate commands. Consider a dispatch table or strategy pattern if this is expected to expand.

---

### **[STYLE NOTES]** (Minor)

**[utils.py, Lines 116-135] 📝 Docstring Redundancy**

The docstring for `get_default_cli_tools()` lists the tools twice - once in the description and once in the Returns section. The code is self-documenting; the docstring adds noise.

---

## **VERDICT:**
❌ **Needs rework** - The code duplication is a fundamental design issue that must be addressed first.

## **KEY INSIGHT:**
The delegate tool formatting logic should live in ONE place, probably as a method on `DelegateAction` itself or a dedicated formatter module. The current approach of copy-pasting the same logic across streaming, non-streaming, and visualization code paths is a maintenance nightmare waiting to happen.

---

*P.S. - The actual functionality looks correct. It's the architecture that needs work. Good code isn't just about making it work - it's about making it maintainable.*

malhotra5:
@openhands if a conversation didn't contain the delegate tool, we cannot resume the conversation with the delegate tool as well


please make sure that we only pass the tools that the resuming conversation contains when creating the agent

malhotra5:
@openhands /codereview-roasted. post the review back as a github comment

malhotra5:
I've done multiple rounds of code review roasted and I'm happy with the implementation so far; marking as ready for human review

pyproject.toml
```diff
@@ -138,6 +138,6 @@ typeCheckingMode = "standard"
 
 # USE EXACT COMMIT FROM AGENT-SDK (PR #1368 - Add condense method to conversation class)
 
-# [tool.uv.sources]
-# openhands-sdk = { git = "https://github.com/OpenHands/software-agent-sdk.git", subdirectory = "openhands-sdk", rev = "5ee81068fd16bb84d9a611e9af44825478155e15" }
-# openhands-tools = { git = "https://github.com/OpenHands/software-agent-sdk.git", subdirectory = "openhands-tools", rev = "5ee81068fd16bb84d9a611e9af44825478155e15" }
+[tool.uv.sources]
+openhands-sdk = { git = "https://github.com/OpenHands/software-agent-sdk.git", subdirectory = "openhands-sdk", rev = "c76e5bd220ea9eb95a9d351313bdb4e533c9da6e" }
```
aaron-seq: **Critical: CI/CD Failure**

This change is causing the `Check Package Versions` workflow to fail. The CI reports:

```
FAILED: Found dependencies using 'rev' fields:
✗ openhands-sdk in [tool.uv.sources] uses rev='c76e5bd220ea9eb95a9d351313bdb4e533c9da6e' (NOT ALLOWED)
✗ openhands-tools in [tool.uv.sources] uses rev='c76e5bd220ea9eb95a9d351313bdb4e533c9da6e' (NOT ALLOWED)
```

According to the CI check, using `rev` fields for dependencies is not allowed. The error message suggests:
> Please use versioned releases instead, e.g.:
> my-package = "1.0.0"
>
> For uv sources, use tagged versions:
> my-package = { git = "https://github.com/user/repo.git", tag = "v1.0.0" }

You need to either:
1. Use a tagged version instead of a commit hash (preferred)
2. Wait for a proper release of openhands-sdk and openhands-tools and use the versioned release
3. Check with maintainers if there's an exception process for this specific use case

openhands_cli/stores/agent_store.py
```diff
@@ -13,20 +15,61 @@
...
+        the conversation doesn't exist or can't be read
+    """
+    conversation_dir = os.path.join(CONVERSATIONS_DIR, conversation_id)
+    base_state_path = os.path.join(conversation_dir, BASE_STATE)
+
+    if not os.path.exists(base_state_path):
+        return None
+
+    try:
+        with open(base_state_path) as f:
```
aaron-seq, line 63: **Missing encoding parameter**

The `open()` call should explicitly specify `encoding='utf-8'` to avoid potential encoding issues across different platforms and locales.

Recommended fix:
```python
with open(base_state_path, encoding='utf-8') as f:
```

This is a best practice to ensure consistent behavior regardless of the system's default encoding.

openhands_cli/shared/delegate_formatter.py
```diff
@@ -0,0 +1,71 @@
...
+            if agent_type and agent_type != "default":
+                agents_info.append(f"{agent_id} ({agent_type})")
+            else:
+                logger.warning("Length of IDs did not match agent types")
+                agents_info.append(agent_id)
+        agents_str = ", ".join(agents_info)
+    else:
+        agents_str = ", ".join(ids)
+
+    return f"Spawning {len(ids)} sub-agent(s): {agents_str}"
```
aaron-seq, line 64: **Minor: Improve pluralization**

Using `(s)` for pluralization is not ideal for user-facing strings. Consider using proper conditional pluralization:

```python
agent_word = "sub-agent" if len(ids) == 1 else "sub-agents"
return f"Spawning {len(ids)} {agent_word}: {agents_str}"
```

This provides better UX by showing "1 sub-agent" vs "2 sub-agents" instead of "1 sub-agent(s)" or "2 sub-agent(s)".

aaron-seq (changes requested):
## Code Review Summary

Thank you for this comprehensive PR adding delegate tool support for spawning sub-agents. The implementation is well-structured with good test coverage. However, there are issues that need to be addressed before merging.

### Critical Issues (Must Fix)

**1. CI/CD Failure - Blocking**
The `Check Package Versions` workflow is failing due to the use of `rev` fields in `pyproject.toml`. This must be resolved before the PR can be merged. See my inline comment for details and suggested fixes.

### Code Quality Issues

**2. Missing encoding parameter**
The `open()` call in `agent_store.py` line 57 should explicitly specify `encoding='utf-8'` to prevent encoding issues across platforms.

**3. Pluralization improvement**
The `sub-agent(s)` format in user-facing strings could be improved with proper conditional pluralization for better UX.

### Positive Aspects

✅ Excellent test coverage including unit tests, integration tests, and streaming tests
✅ Clean separation of concerns with the centralized `delegate_formatter.py` module
✅ Proper handling of conversation tool preservation when resuming
✅ Well-documented code with clear docstrings
✅ Follows the repository's coding standards and patterns

### Recommendations

1. **Priority**: Fix the CI/CD failure by updating the dependency specification in `pyproject.toml`
2. Add `encoding='utf-8'` to file operations
3. Consider improving the pluralization for better user experience
4. Once these are addressed, this PR will be ready to merge

### Testing

The PR includes comprehensive tests that cover:
- Delegate tool streaming behavior
- Tool preservation across conversation resumption
- Both TUI and ACP delegate functionality
- Edge cases and error handling

Overall, this is a solid implementation that just needs the CI/CD issue resolved.

openhands_cli/shared/delegate_formatter.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def _format_delegate_tasks_title(tasks: dict[str, Any] | None) -> str:
+    """Format title for delegate command."""
+    if not tasks:
+        return "Delegating tasks"
+    return f"Delegating {len(tasks)} task(s) to: {', '.join(tasks.keys())}"
```
xingyaoww, line 71: Shouldn't we be able to use ".summary" from the ActionEvent for these?

openhands_cli/tui/widgets/richlog_visualizer.py
```diff
@@ -459,29 +580,62 @@ def _create_event_widget(self, event: Event) -> "Widget | None":
             ):
                 return None
             # Display messages as markdown for proper rendering
-            widget = Markdown(str(content))
+            # In delegation context, prefix messages with agent info (underlined)
+            message_content = str(content)
+            if self._name and event.llm_message:
+                agent_name = self._get_formatted_agent_name()
+                # Use markdown underline syntax for agent names
+                agent_underlined = f"__{agent_name}__"
```
malhotra5: @openhands can you remove all the special logic in place for underlining in markdown syntax

malhotra5:
Delegate events are also visualized now

image

openhands_cli/shared/delegate_formatter.py
malhotra5: Using summary now, and this is the fallback title

malhotra5:
@openhands /codereview-roasted. post back the review as a github comment

malhotra5:
## 🔥 Code Review: Delegate Tool Support (Round 2)

**Taste Rating:** 🟡 **Acceptable** - It works, but the codebase is slowly accumulating the kind of cruft that makes future maintainers weep.

---

**[openhands_cli/stores/agent_store.py, Line 57]** 🔒 **Encoding Omission (STILL HERE)**
```python
with open(base_state_path) as f:
```
I flagged this in the last review. It's *still* here. We're in 2026 and we're *still* playing Russian roulette with file encodings? Always specify `encoding='utf-8'`. This is the kind of "it works on my machine" code that explodes spectacularly in production when someone's conversation state contains a non-ASCII character.

**[openhands_cli/tui/widgets/richlog_visualizer.py, Lines 577-595]** 🤯 **Logic Spaghetti**
```python
if (
    self._skip_user_messages
    and event.llm_message
    and event.llm_message.role == "user"
    and not event.sender
):
    return None

if not event.sender or event.llm_message or self._name:
    return None
```
What in the actual... Let me trace this logic: If we have a user message without a sender, skip it. Then, if we DON'T have a sender OR we DO have an llm_message OR we have a name... skip it too? This is the kind of conditional logic that makes me want to throw my keyboard. The second condition will return None for almost everything. I suspect this is a bug, not a feature.

---

**[openhands_cli/acp_impl/events/tool_state.py, Lines 110-113, 222-227]** 📈 **The Special Case Cancer Spreads**
```python
if self.tool_name == "delegate":
    return "other"
```
Every. Single. Tool. Gets. Its. Own. If. Statement. This is the textbook example of what I call "bad taste" - you're not solving the problem, you're just adding more band-aids. When you add the 10th tool, you'll have 10 if statements scattered across 3 files. The solution is a registry pattern or a method on the tool itself. Fix the data structure, not the code.

**[openhands_cli/shared/delegate_formatter.py, Lines 55-58]** 🤔 **Warning That Lies**
```python
if agent_type and agent_type != "default":
    agents_info.append(f"{agent_id} ({agent_type})")
else:
    logger.warning("Length of IDs did not match agent types")
    agents_info.append(agent_id)
```
This warning message is a lie. It fires when `agent_type` is `None` OR when it equals `"default"`. Neither of those conditions means "length mismatch". You're going to confuse the hell out of someone debugging this at 2 AM. Either fix the message or fix the logic.

**[openhands_cli/tui/widgets/richlog_visualizer.py, Lines 148-175]** 🎭 **Over-Engineered Name Formatting**
```python
@staticmethod
def _format_agent_name(name: str) -> str:
    # If already has spaces, assume it's already formatted
    if " " in name:
        return name
    # Handle snake_case...
    if "_" in name:
        return name.replace("_", " ").title()
    # Handle camelCase/PascalCase...
    spaced = re.sub(r"(?<!^)(?=[A-Z])", " ", name)
    return spaced.title()
```
You've written 15 lines of code to handle 3 different naming conventions. What happens when someone uses `kebab-case`? Or `SCREAMING_SNAKE_CASE`? Or `mixedCase_with_underscores`? This is solving an imaginary problem with incomplete coverage. Just use the name as-is, or require a display_name field.

**[openhands_cli/tui/widgets/richlog_visualizer.py, _create_event_widget method]** 📏 **Function Obesity**
This method is now ~100 lines with 8+ different code paths. It's doing too much. Each event type deserves its own handler method. The current structure is a maintenance nightmare waiting to happen.

---

**[pyproject.toml, Lines 19-20]** 📌 **Git Commit Pinning**
```toml
"openhands-sdk @ git+https://github.com/OpenHands/software-agent-sdk.git@3dd519e6b37aa745fe2b95be93cc300244d46286"
```
Pinning to a git commit hash instead of a versioned release is fragile. When that commit gets garbage collected or the repo gets reorganized, your build breaks. I understand this might be temporary for development, but don't forget to switch back to proper versioned dependencies before this goes to production.

**[tests/shared/test_delegate_formatter.py]** ✅ **Good Test Coverage**
I'll give credit where it's due - the test coverage for the delegate formatter is solid. Parameterized tests, edge cases covered. This is how tests should be written.

---

### **VERDICT:**
✅ **Worth merging** (with reservations) - The core functionality is sound and well-tested. However:
1. Fix the encoding issue (seriously, this is the second time)
2. Investigate the logic bug in `_create_event_widget`
3. Consider refactoring the special-case pattern before it metastasizes further

---

### **KEY INSIGHT:**
The delegate tool implementation itself is fine. The problem is that each new tool is adding more special cases to shared code paths instead of using polymorphism or a registry pattern. You're building technical debt that compounds with every new feature. Fix the architecture, not just the symptoms.

tests/tui/widgets/test_richlog_visualizer.py
```diff
@@ -815,3 +815,137 @@ def on_mount(self):
...
+
+        assert sub_vis._skip_user_messages is True
+
+
+class TestMessageEventDelegation:
+    """Tests for MessageEvent handling in delegation context."""
+
+    def test_message_event_skipped_without_delegation_context(self, visualizer):
+        """MessageEvent returns None when not in delegation context."""
+        from openhands.sdk import Message
```
enyst, line 858: I honestly am starting to feel that inline imports, at least in tests, don't matter.

Outside tests, it's possible that if we continue to keep the rule, inline imports smell a bit, and then we can look closer at the source of that smell and see if there's something to fix.

On the other hand, even outside tests... we could also let them be, and then occasionally e.g. bi-weekly run a sweep. Maybe... WDYT?

malhotra5: Yep I think inline-imports can be cleaned up by the bi-weekly run

Parts of the CLI do inline import on purpose though, as it lazy loads specific modules as needed for a fast startup 

E.g when running TUI we don't import any ACP related logic

enyst: Totally, sometimes they're the right thing to do. I think the agent is smart enough to avoid messing with some - it already has.

PR #333, opened 2026-01-10
feat(tui): conversation history side panel

pashkash:
### What changed
- Added an in-app Conversation History side panel toggled via /history and Ctrl+H.
- Implemented conversation switching from the panel with a persistent Loading toast that appears before loading and is dismissed after completion.
- Moved switch loading to a thread worker to keep the UI responsive.
- Improved History UX:
  - Title derived from the first user message (up to 100 chars).
  - Full conversation id shown as secondary info.
  - In-place highlighting for current and selected items (no list reload).
  - “New conversation” entry updates immediately after the first message while the panel is open.
- Added incremental loading in the history list: shows 10 most recent by default, “… More” loads 10 more.

### Cloud support
- If the user is logged in to OpenHands Cloud (token present), the panel also lists cloud conversations.
- Selecting a cloud conversation performs a best-effort resume by opening the cloud conversation URL in the browser.

### Test status
- make lint: ✅
- uv run pytest: ✅ (853 passed)

PR #355, opened 2026-01-14
feat(tui): rehydrate UI with conversation history on resume (#204)

### LINKED ISSUE #204: Rehydrate conversation pane with event history for resumed conversation
When resuming conversations none of the prior events from the conversation trajectory are visible in the UI. We should rehydrate the UI with the old events on setup for resumed conversations.

pashkash:
## Summary

Closes #204

When resuming a conversation, the UI now displays the conversation history so users can see prior events. This implements paginated history loading for better UX with long conversations.

### New module: `openhands_cli/tui/core/history_replay.py`
- `HistoryReplayState` dataclass for tracking pagination state
- `get_initial_page()` - returns the last N events on initial load
- `get_next_page()` - returns older events on scroll-up
- `replay_events_to_visualizer()` - replays events through the visualizer

### Integration in `textual_app.py`
- On conversation switch, load and display the last 5 events
- On scroll-up (when near top), load 5 more older events
- Events are prepended to the display (after splash widgets)
- Shows notification when loading more history

### Tests
- 16 comprehensive tests for pagination logic in `tests/tui/core/test_history_replay.py`

## How it works

1. When switching to an existing conversation, `_switch_to_conversation_thread` creates a `HistoryReplayState` from the conversation's events
2. The initial page (last 5 events) is replayed through the `ConversationVisualizer`
3. When user scrolls up and reaches near the top (`scroll_y < 50`), more events are loaded
4. Events are rendered using the same visualizer that handles live events, ensuring consistent appearance

## Testing

```bash
# Run the new tests
uv run pytest tests/tui/core/test_history_replay.py -v

# All tests pass
make test
make lint
```

## Screenshots

N/A - This is backend/logic change. The UI renders the same event widgets as before, just populated with historical data on resume.

### DIFF
--- a/openhands_cli/tui/textual_app.py
+++ b/openhands_cli/tui/textual_app.py
@@ -73,6 +73,7 @@ class OpenHandsApp(CollapsibleNavigationMixin, App):
         ("ctrl+h", "toggle_history", "Toggle history panel"),
         Binding("ctrl+u", "load_more_history", "Load more history", show=False),
         ("ctrl+j", "submit_textarea", "Submit multi-line input"),
+        ("ctrl+x", "open_external_editor", "Open external editor"),
         ("escape", "pause_conversation", "Pause the conversation"),
         ("ctrl+q", "request_quit", "Quit the application"),
         ("ctrl+c", "request_quit", "Quit the application"),
@@ -750,6 +751,12 @@ class OpenHandsApp(CollapsibleNavigationMixin, App):
             self.conversation_runner.pause_runner_without_blocking()
         else:
             self.notify(message="No running conversation to pause", severity="error")
+
+    def action_open_external_editor(self) -> None:
+        """Forward open external editor action to the input field."""
+        # Debug: notify that the main app action was triggered
+        self.notify("Main app CTRL+X action triggered", severity="information")
+        self.input_field.action_open_external_editor()
 
     def _handle_confirm_command(self) -> None:
         """Handle the /confirm command to show confirmation settings modal."""
--- a/openhands_cli/tui/widgets/input_field.py
+++ b/openhands_cli/tui/widgets/input_field.py
@@ -1,3 +1,8 @@
+import os
+import shutil
+import subprocess
+import tempfile
+from pathlib import Path
 from typing import TYPE_CHECKING, ClassVar, cast
 
 from textual import on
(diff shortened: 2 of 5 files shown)


openhands_cli/tui/core/history_replay.py
```diff
@@ -0,0 +1,126 @@
+"""History replay with pagination for conversation UI.
```
enyst, line 1: I still wish we didn't use the term "replay" for simply display / resend, but I think I lost this one. 😅 

My reason was that in OpenHands, _replay_ used to mean [actually re-executing the agent's history](https://github.com/OpenHands/OpenHands/issues/6049) in a new environment.

(With the local <-> remote fun, we are maybe likely IMHO to need _some_ term for actually re-executing, since we seem to have repurposed this one... I don't know what that could be... re-... something 🤷, dunno, maybe not important atm)

openhands_cli/tui/textual_app.py
```diff
@@ -743,6 +881,313 @@ def _handle_new_command(self) -> None:
...
+            group="switch_conversation",
+            exclusive=True,
+            thread=True,
+            exit_on_error=False,
+        )
+
+    def _switch_to_conversation_with_stop(self, target_id: uuid.UUID) -> None:
+        """Switch conversations, pausing the current run if needed."""
+        # Show a persistent loading notification (we will dismiss it manually).
+        self._show_switch_loading_notification()
```
enyst, line 974: Switching conversation seems like a different thing than just display conversation history on resume. I'm a bit surprised it made it in the same PR, or am I missing something? Not a big deal, just wondering.

openhands_cli/tui/core/history_replay.py
pashkash: yes, we just show the events here and we dont replay them (but after we send next massage - new event will be emitted based on the current state of the convo if it dont mistake). so lets stick with "display" as you suggested;

openhands_cli/tui/textual_app.py
pashkash: malhotra5 pushed me a bit to shy less by suggesting making new files and be free to think about the structure and entities; by his suggestion i [created](https://github.com/OpenHands/OpenHands-CLI/pull/333/files#diff-9a5a13e0539c2235e92dda8527cf5d2492f0b7ae3333d80bb018c92c3df1329e) a conversation switcher. 
and this branch should be adjusted with these changes; i think i'll mark this branch as "draft" as you suggested before; and after the history pane feature will be merged i will back to this feature branch, update it with the latest changes and ask you for review; until that pls take a look on the [history pane](https://github.com/OpenHands/OpenHands-CLI/pull/333/files#diff-9a5a13e0539c2235e92dda8527cf5d2492f0b7ae3333d80bb018c92c3df1329e) if you have time; ty a lot!

openhands_cli/tui/core/history_replay.py
enyst: Not a big deal either way, thank you!

PR #361, opened 2026-01-16
Fix cloud login to respect llm_base_url from settings

enyst:
### Problem
`openhands login` fetches `/api/settings` from OpenHands Cloud, but ignores `llm_base_url` when creating the local Agent config.

This makes the CLI *always* display/save the default proxy URL (`https://llm-proxy.app.all-hands.dev/`), even when the user configured a custom base URL (e.g. the `eval` proxy) in Cloud settings.

### Fix
- Resolve `base_url` from `settings['llm_base_url']` (fallback to the default proxy URL).
- Use the resolved base URL in:
  - the overwrite-consent prompt
  - the saved Agent LLM config (agent + condenser)
- Make `AgentStore.create_and_save_from_settings()` default to `settings['llm_base_url']` when `base_url` is not explicitly provided.

### Tests
- `python -m pytest`

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@fix/cloud-login-llm-base-url
```

### DIFF
--- a/openhands_cli/auth/api_client.py
+++ b/openhands_cli/auth/api_client.py
@@ -9,6 +9,7 @@ from openhands_cli.auth.http_client import AuthHttpError, BaseHttpClient
 from openhands_cli.auth.utils import _p
 from openhands_cli.locations import AGENT_SETTINGS_PATH, PERSISTENCE_DIR
 from openhands_cli.stores import AgentStore
+from openhands_cli.stores.agent_store import resolve_llm_base_url
 from openhands_cli.theme import OPENHANDS_THEME
 
 
@@ -195,10 +196,7 @@ def create_and_save_agent_configuration(
     """
     store = AgentStore()
 
-    settings_base_url = settings.get("llm_base_url")
-    base_url = str(settings_base_url).strip() if settings_base_url is not None else ""
-    if not base_url:
-        base_url = "https://llm-proxy.app.all-hands.dev/"
+    base_url = resolve_llm_base_url(settings)
 
     # First, check if existing configuration exists
     existing_agent = store.load()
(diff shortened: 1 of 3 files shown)


openhands_cli/stores/agent_store.py
```diff
@@ -122,33 +122,40 @@ def create_and_save_from_settings(
         self,
         llm_api_key: str,
         settings: dict[str, Any],
-        base_url: str = "https://llm-proxy.app.all-hands.dev/",
+        base_url: str | None = None,
```
enyst, line 144: Oh, it looks like that's it. We weren't sending it.

openhands_cli/auth/api_client.py
```diff
@@ -195,17 +195,27 @@ def create_and_save_agent_configuration(
     """
     store = AgentStore()
 
+    settings_base_url = settings.get("llm_base_url")
+    base_url = str(settings_base_url).strip() if settings_base_url is not None else ""
+    if not base_url:
+        base_url = "https://llm-proxy.app.all-hands.dev/"
+
```
malhotra5: NIT: I think we can dedup this against the logic inside `create_and_save_from_settings` (for getting the base url from the settings dict)

malhotra5:
Thanks a bunch for this!

openhands_cli/auth/api_client.py
enyst: Done.

And, ha! The agent got access to `gh`, and now it's happily using it to mark comments as solved. (I didn't tell it that)

malhotra5 (approved):
Thank you!

enyst:
Live-tested this change via a local `uv build` wheel +:

```bash
uvx --isolated --from dist/openhands-1.9.0-py3-none-any.whl openhands login --server-url https://app.all-hands.dev
```

Result: CLI now saves/displays `llm_base_url` from Cloud settings correctly (saw `Base URL: https://llm-proxy.eval.all-hands.dev/`).

PR #371, opened 2026-01-18
Clean-up dead-code in `openhands-cli/argparsers`

VascoSch92:
While looking through the repo, I discovered that the `openhands-cli/argparsers` directory contains two utility files: `util.py `and `utils.py`. It appears that the latter contained dead code. Indeed, it wasn't being used, and its contents were already present in `util.py`. This happened probably because of a rebase.

 I have deleted the `utils.py` file to clean this up.


malhotra5 (approved):
Thank you!

PR #374, opened 2026-01-19
docs(CONTRIBUTING): Correct install pre-commit hooks command 

utkarsh-in:
docs(CONTRIBUTING): Correct install pre-commit hooks command in AGENTS.md


enyst:
@openhands What is the difference between the two? WDYT, if you run one (separately in a worktree) and then the other, which is correct / better and why?

enyst:
Thank you @utkarsh-in ! I asked the OpenHands agent for its opinion. Sorry, I wasn't aware of the discrepancy here. 😅

malhotra5 (approved):
Thank you!

PR #277, opened 2025-12-31
Exp: remote acp conversations 

malhotra5:
The ability to have a conversation with an openhands agent via ACP. The conversation should take place in a remote runtime in the cloud, and only supports V1 conversations. 

Clarifications required
- Do we want to support cloning the repository and checking out a branch; how would we identify the repo?

QA checklist that I've verified in Zed. You can use the following config to test this PR

```json
"openhands_remote_acp": {
      "type": "custom",
      "command": "uvx",
      "args": [
        "--python",
        "3.12",
        "git+https://github.com/OpenHands/OpenHands-CLI.git@remote-acp-convo",
        "acp",
        "--cloud",
      ],
      "env": {},
},
```

- [X] Request to oauth if not logged in
- [X] Start fresh conversation
- [X] Resume existing conversation (started via ACP, or existing V1 conversation from cloud)
- [X] Confirmation mode
- [X] Sending a prompt to a conversation that has been open for a while resumes the sandbox 

#### Oauth if not authenticated
https://github.com/user-attachments/assets/f169ac81-32f9-4020-8499-69181da67710

#### Starting fresh convo + confirmation mode

https://github.com/user-attachments/assets/f01022d0-61cd-4d1e-9876-c52eb559f8d4

#### Resuming existing convo
https://github.com/user-attachments/assets/c018a1d3-4ca9-42a5-8d93-b48e15ca19a3

#### Resume a closed sandbox for conversations that have been open for a while 
https://github.com/user-attachments/assets/6fc1e2e7-487a-46a0-ac4f-328e56e1d71b

Related to #269

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@remote-acp-convo
```

PR #388, opened 2026-01-22
(Hotfix): fix headless mode to work with env overrides + disable critic in headless mode

malhotra5:
## Summary

This PR fixes an issue where `--override-with-envs` flag didn't work with headless mode.

### The Problem

When using headless mode (`--headless`) with environment variable overrides (`--override-with-envs` and `LLM_API_KEY` set), the CLI would fail with:

```
Headless mode requires existing settings.
Please run: openhands to configure your settings before using --headless.
```

This happened because `AgentStore.load()` only returned an agent if a settings file existed, ignoring the fact that valid configuration could be constructed from environment variables alone.

### The Fix

Modified `AgentStore.load()` in `openhands_cli/stores/agent_store.py` to:

1. **When no settings file exists**, check if `--override-with-envs` is enabled and `LLM_API_KEY` environment variable is set
2. If so, **create an agent from environment variables** using:
   - `LLM_API_KEY` (required)
   - `LLM_BASE_URL` (optional, defaults to `https://llm-proxy.app.all-hands.dev/`)
   - `LLM_MODEL` (optional, defaults to `claude-sonnet-4-5-20250929`)

Added a new private method `_create_agent_from_env_overrides()` that handles creating the agent configuration from environment variables.

### Changes Made

1. **`openhands_cli/stores/agent_store.py`**:
   - Refactored `load()` method to handle `FileNotFoundError` and try to create agent from env vars
   - Added `_create_agent_from_env_overrides()` method

2. **`openhands_cli/tui/modals/settings/settings_screen.py`**:
   - Updated docstring for `is_initial_setup_required()` to document the new behavior

### Testing

All 72 relevant tests pass, including:
- Headless mode tests
- Environment variable override tests
- Settings screen tests

### Usage

Now you can run headless mode without pre-existing settings:

```bash
LLM_API_KEY=your-key openhands --headless --override-with-envs -t "Fix bug"
```

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@fix-headless-env-overrides
```

### DIFF
--- a/openhands_cli/stores/agent_store.py
+++ b/openhands_cli/stores/agent_store.py
@@ -310,31 +310,35 @@ class AgentStore:
         env_overrides_enabled: bool = False,
         critic_disabled: bool = False,
     ) -> Agent | None:
-        """Load agent from disk, or create from env vars if enabled.
+        """Load an Agent and apply runtime configuration.
 
-        This is the main entry point for the CLI to get a configured agent.
-        It orchestrates the policy of preferring environment variables when
-        env_overrides_enabled is True.
+        Invariant:
+        - If a persisted agent exists:
+            * Load it from disk.
+            * Apply any env overrides that are present (even partial).
+        - If no persisted agent exists:
+            * Require a full env spec (LLM_API_KEY + LLM_MODEL) to create
+                a default Agent.
+            * Otherwise, raise an error.
+
+        Runtime configuration (tools, context, MCP, metadata, critic) is
+        always applied last.
 
         Args:
-            session_id: Optional session ID for metadata tracking and tool restoration.
-            env_overrides_enabled: If True, prefer creating agent from environment
-                variables over loading from disk. This ensures env vars take
-                precedence when the user explicitly requests it. If all required
-                env vars (LLM_API_KEY, LLM_MODEL) are set, creates agent entirely
-                from env vars. Otherwise, loads from disk and applies partial
-                env var overrides.
-            critic_disabled: If True, critic functionality will be disabled
-                (e.g., for headless mode).
+            session_id: Optional session ID used for tool restoration and
+                LLM metadata tagging.
+            env_overrides_enabled: Whether env overrides are enabled.
+            critic_disabled: If True, do not configure a critic.
 
         Returns:
-            Configured Agent instance, or None if no configuration exists and
-            env_overrides_enabled is False.
+            A fully configured Agent, or None if no persisted agent exists and
+            env overrides are disabled.
 
         Raises:
-            MissingEnvironmentVariablesError: If env_overrides_enabled is True,
-                no disk config exists, and required environment variables are missing.
+            MissingEnvironmentVariablesError: If no persisted agent exists and
+                required env variables are missing.
         """
+
         agent = self.load_from_disk()
         overrides = LLMEnvOverrides.from_env(enabled=env_overrides_enabled)
 


malhotra5:
QA works as expected

```
(openhands) rohitmalhotra@RohitMalhotra-MacBook-Pro OpenHands-CLI % uv run openhands --headless --override-with-envs -t "echo hello world" 
      Built openhands @ file:///Users/rohitmalhotra/Documents/OpenHands-CLI
Uninstalled 1 package in 2ms
Installed 1 package in 2ms
Initializing agent...
✓ Agent initialized with model: litellm_proxy/prod/claude-opus-4-5-20251101

─────────────────────────────────────────────────────────────────────────────────── CONVERSATION SUMMARY ────────────────────────────────────────────────────────────────────────────────────
Number of agent messages: 3
Last message sent by the agent:
╭─ Agent ─────╮
│ hello world │
╰─────────────╯
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Goodbye! 👋
Conversation ID: 599bdc934ddc49e5850096a772413521
Hint: run openhands --resume 599bdc93-4ddc-49e5-8500-96a772413521 to resume this conversation.
(openhands) rohitmalhotra@RohitMalhotra-MacBook-Pro OpenHands-CLI % 
```

openhands_cli/stores/agent_store.py
```diff
@@ -265,112 +289,164 @@ def __init__(self) -> None:
...
+        api_key = env_overrides.api_key.get_secret_value()
+        base_url = env_overrides.base_url or DEFAULT_LLM_BASE_URL
+        model = env_overrides.model or "claude-sonnet-4-5-20250929"
+
+        llm = LLM(
+            model=model,
+            api_key=api_key,
+            base_url=base_url,
+            usage_id="agent",
+        )
```
malhotra5: @openhands notice the fallback values for api_key, base_url, and model

don't create these fallbacks; at the very least we should have api key and model, and if not raise a known error with a human readable message

examine the entrypoints and ensure that when envs are passed properly a proper error message is plumbed back

malhotra5:
QA - missing envs also have an explainable error message

```
rohitmalhotra@RohitMalhotra-MacBook-Pro OpenHands-CLI % uv run openhands --headless --override-with-envs -t "echo hello world"
Error: Missing required environment variable(s): LLM_API_KEY
When using --override-with-envs, you must set:
  - LLM_API_KEY: Your LLM API key
  - LLM_MODEL: The model to use (e.g., claude-sonnet-4-5-20250929)
rohitmalhotra@RohitMalhotra-MacBook-Pro OpenHands-CLI % 
```

enyst:
Sorry, I know this is not from this PR, but FWIW I think maybe this should be None?
> LLM_BASE_URL (optional, defaults to https://llm-proxy.app.all-hands.dev/)

It's unexpected that people need to set it to nothing, if I understand correctly.

enyst:
@openhands Do a /codereview-roasted on this PR.

Feel free to investigate deeply to understand the behavior.

malhotra5:
OH nice catch, you're right that default should not be set

malhotra5:
@enyst it seems the review didn't it back to github as a comment

openhands_cli/stores/agent_store.py
```diff
@@ -265,112 +308,180 @@ def __init__(self) -> None:
...
+                    llm_type="agent",
+                    session_id=session_id,
+                )
+            }
+        if llm_update:
+            updated_llm = updated_llm.model_copy(update=llm_update)
+
+        # Always create a fresh condenser with current defaults if condensation
+        # is enabled. This ensures users get the latest condenser settings
+        # (e.g., max_size, keep_first) without needing to reconfigure.
```
enyst: @openhands Verify this behavior please.

Make the project, then run it. You can install yourself tmux and gh if you don't have them already, and whatever you need.

And verify: if the user *has* agent_settings.json, already on disk (look it up to understand when / how it's created), then are their condenser vars overridden *with defaults*?

You have env vars you can use, you can use tmux and snapshot to see.

If yes, make an issue on it, and then make a new branch from main and fix it, then open a PR where we propose the fix because the defaults shouldn't actually override user settings. Do not commit to this branch, it's a different issue.

enyst:
🔴 **Taste rating: Needs improvement**  
It “works”, but the solution adds global state, duplicates configuration logic, and spreads headless behavior across unrelated layers. This is the kind of fix that will age like milk the next time someone tweaks startup flow.

---

## [CRITICAL ISSUES] (must fix / high risk)

### 1) Global flags as configuration transport
- **File:** `openhands_cli/stores/agent_store.py` (`set_env_overrides_enabled()`, `set_critic_disabled()`)
- **What’s wrong:** these are process-global toggles used as hidden inputs to agent construction.
- **Why it’s bad:** order-dependent behavior, test pollution, and future “why is this disabled?” debugging. Even in a CLI, this runs in-process and can be imported/used in ways you don’t control.
- **Concrete improvement:** pass explicit options through an object or function args (e.g., `AgentStore(load_options=...)` or `AgentStore.load(..., env_overrides_enabled=..., critic_disabled=...)`). At minimum, make lifecycle explicit and reset deterministically per invocation.

### 2) `AgentStore.load()` now has two meanings
- **File:** `openhands_cli/stores/agent_store.py` (`load()`)
- **What’s wrong:** `load()` now means “load persisted agent **or** synthesize one from env vars if missing”.
- **Why it’s bad:** a method named `load()` should load stored state. This hides policy behind a `FileNotFoundError` path.
- **Concrete improvement:** split responsibilities:
  - `load_from_disk()` → only disk
  - `create_from_env_overrides()` → only env
  - `load_or_create()` → explicit orchestration (policy chosen by the CLI)

### 3) Critic disabling is bolted on in the wrong place
- **File:** `openhands_cli/entrypoint.py` (headless sets `set_critic_disabled(True)`)
- **What’s wrong:** UI/runtime mode (`--headless`) is reaching into agent construction via a global.
- **Why it’s bad:** cross-layer coupling with no contract; any future startup path that bypasses `entrypoint.main()` won’t get consistent behavior.
- **Concrete improvement:** encode “headless implies no critic” in the headless runner / agent factory by passing options explicitly, not through global store state.

---

## [IMPROVEMENT OPPORTUNITIES] (should fix / “bad taste”)

### 4) Default base URL logic is duplicated / surprising
- **File:** `openhands_cli/stores/agent_store.py`
- **What’s wrong:** env-path creation re-implements base URL defaulting (`env_overrides.base_url or DEFAULT_LLM_BASE_URL`) instead of reusing `resolve_llm_base_url()`.
- **Why it’s bad:** multiple sources of truth and inconsistent policy drift.
- **Concrete improvement:** unify base URL resolution in one place. Also, per reviewer feedback: defaulting to `https://llm-proxy.app...` may be a product choice, but it’s surprising if the user expected “unset means None”.

### 5) Error handling depends on whether the settings file exists
- **Files:** `agent_store.py` + `entrypoint.py`
- **What’s wrong:** `MissingEnvironmentVariablesError` is only triggered on the “no settings file” path. If a settings file exists (even stale/bad), you may get a different, less helpful failure.
- **Concrete improvement:** validate required env vars at the boundary where `--override-with-envs` is accepted (arg parsing / CLI startup), not only deep inside `AgentStore.load()`.

### 6) Tests are now babysitting globals
- **File:** `tests/tui/test_headless_mode.py`
- **What’s wrong:** tests have to manually reset global flags (“clean up”).
- **Why it’s bad:** this is the design smell proving the design problem.
- **Concrete improvement:** remove globals or scope them to an instance (so tests don’t leak state across cases).

### 7) Policy mismatch: is `LLM_MODEL` required or defaulted?
- There’s some inconsistency between messaging and behavior: current code requires `LLM_MODEL` in env mode (good for explicitness), but earlier text implied defaulting. Pick one policy and make messaging + behavior consistent.

---

## VERDICT
✅ Fixes a real bug, but the implementation introduces avoidable complexity.  
❌ If we want this to be maintainable, remove hidden global toggles and make “create agent from env” an explicit, first-class pathway.

**Key insight:** the bug fix is legitimate, but the current shape uses **hidden global state + implicit fallback construction**, which guarantees future “why is it doing that?” debugging sessions.

malhotra5:
@openhands /codereview-roasted. post back there review as a github comment

malhotra5:
## 🔥 /codereview-roasted

🟢 **Taste Rating: Good taste** - This is actually a well-designed refactoring that removes problematic global state in favor of explicit parameter passing.

---

## [CRITICAL ISSUES] (none found)

Surprisingly, this PR doesn't have critical issues. The previous reviewer's concerns about "global flags as configuration transport" were actually describing the **old** code that this PR **removes**. The `set_env_overrides_enabled()` global flag is gone - that's the fix, not the problem.

---

## [IMPROVEMENT OPPORTUNITIES]

### 1) Parameter threading verbosity
- **Files:** `entrypoint.py` → `textual_app.py` → `conversation_runner.py` → `setup.py` → `agent_store.py`
- **What's happening:** `env_overrides_enabled` and `critic_disabled` are threaded through 5+ layers of function calls.
- **Assessment:** This is verbose but **correct**. Explicit parameter passing is better than global state. The old code used `set_env_overrides_enabled(True)` which mutated global state - that's the anti-pattern this PR fixes.
- **Suggestion:** If this becomes unwieldy, consider a configuration context object, but for two boolean flags, explicit parameters are fine.

### 2) Inconsistent `is_initial_setup_required()` call
- **File:** `openhands_cli/tui/modals/settings/settings_screen.py`, Line 79
- **What's wrong:** `self.is_initial_setup = SettingsScreen.is_initial_setup_required()` is called without `env_overrides_enabled` parameter.
- **Impact:** The settings screen always checks without considering env overrides, which could cause inconsistent behavior if someone opens settings while using `--override-with-envs`.
- **Suggestion:** Pass `env_overrides_enabled` to the `SettingsScreen` constructor and use it here.

### 3) `_ensure_agent()` assertion after validation
- **File:** `openhands_cli/stores/agent_store.py`, Lines 282-283
```python
assert overrides.api_key is not None
assert overrides.model is not None
```
- **What's wrong:** These assertions are redundant after `require_for_headless()` which already raises if these are None.
- **Suggestion:** Remove the assertions or convert to a comment explaining the invariant.

---

## [STYLE NOTES]

### 1) Good: Explicit factory method
- **File:** `agent_store.py`, `LLMEnvOverrides.from_env(enabled=True)`
- **Why it's good:** Replacing the magic `@model_validator` that read global state with an explicit factory method is the right call. No more "spooky action at a distance."

### 2) Good: Clear error messages
- **File:** `agent_store.py`, `MissingEnvironmentVariablesError`
- **Why it's good:** The error message tells users exactly what they need to do:
```
Missing required environment variable(s): LLM_API_KEY, LLM_MODEL
When using --override-with-envs, you must set:
  - LLM_API_KEY: Your LLM API key
  - LLM_MODEL: The model to use (e.g., claude-sonnet-4-5-20250929)
```

### 3) Good: Separation of concerns
- `load_from_disk()` - Pure I/O, no policy
- `load_or_create()` - Policy logic with explicit parameters
- `_apply_runtime_config()` - Runtime augmentation (tools, context, MCP, critic)

---

## VERDICT

✅ **Worth merging** - Core logic is sound. The refactoring removes global state and makes configuration flow explicit. The parameter threading is verbose but correct.

---

## KEY INSIGHT

The previous review criticized "global flags as configuration transport" - but that's describing the **old** code that this PR **removes**. The `set_env_overrides_enabled()` function and `_apply_env_overrides` global are gone. This PR replaces implicit global state with explicit parameter passing, which is the right architectural direction.

---

*Note: @enyst's comment about `LLM_BASE_URL` defaulting to the proxy URL is valid - looking at the code, when creating from env vars, `base_url` is passed as `overrides.base_url` which can be `None`. This is actually correct behavior - if not set, it should be `None`, not a hardcoded default. The PR description's mention of a default appears to be outdated documentation.*

malhotra5:
I've QAed the changes and they work as expected

enyst:
@openhands WDYT about this PR? Consider maintainability and easiness to establish / keep / understand invariants.

/codereview-roasted

enyst:
🔴 **Taste rating: Needs improvement**  
It *fixes the bug* and it *removes the worst part (global state)*, but the PR is still doing a lot more than “headless env overrides”, and it mixes policy/UX decisions with persistence/runtime config in ways that make invariants harder to reason about later.

Below is the maintainability / invariants-focused roast.

---

## [CRITICAL ISSUES] (invariants / maintainability risk)

### 1) `load_or_create()` policy is self-contradictory vs docstring
**File:** `openhands_cli/stores/agent_store.py` (around `load_or_create()`)

Docstring says (paraphrased):

- `env_overrides_enabled=True`:
  - **If required env vars are set** → create agent entirely from env
  - **Otherwise** → load from disk and apply partial overrides

But implementation effectively does:

```py
agent = load_from_disk()
overrides = from_env(enabled=env_overrides_enabled)

if env_overrides_enabled:
    agent = _ensure_agent(agent, overrides)  # requires api_key+model if agent is None
    agent = _apply_env_overrides(agent, overrides)
```

So the actual invariant is:
- If disk agent exists: env vars are optional (partial override works)
- If disk agent does *not* exist: api_key+model are required or you raise

That’s totally reasonable — but the docstring currently implies a branch that can’t exist without disk config.

**Suggestion:** tighten the docstring to match the real invariant:
- If disk agent exists: apply any env overrides that exist (even partial)
- If disk agent missing: require full env spec (api_key + model) or error

That invariant is simple and easy to maintain.

---

### 2) Headless-specific behavior leaks into non-headless layers
You added `critic_disabled` plumbing from `entrypoint.py` → `textual_app` → `setup.load_agent_specs` → `AgentStore._apply_runtime_config`.

**Why it’s bad for invariants:** “critic is off” is now a runtime mode flag that any call path must remember to pass correctly, instead of a single policy decision in one place. Miss one call site later and headless starts prompting again.

**Better invariant:** “Critic prompting is controlled by a single interaction/UI mode config, and everything derives from that.”

If you don’t want a config object, at least centralize the decision closer to where critic is created (AgentStore), e.g. `enable_critic = cli_settings.enable_critic and not headless`.

Right now the invariant is “everyone who calls `setup_conversation()` must know whether critic should be disabled”. That’s brittle.

---

### 3) `AgentStore` is doing 3 jobs (persistence + env policy + runtime assembly)
**File:** `openhands_cli/stores/agent_store.py`

You now have:
- disk persistence (`load_from_disk`, `save`)
- env override parsing/validation (`LLMEnvOverrides`, `MissingEnvironmentVariablesError`)
- runtime assembly (tools/context/mcp/condenser/critic metadata)

This increases coupling and makes it harder to preserve invariants like:
- “Agent on disk never contains critic”
- “Env overrides never persist”
- “Resuming a conversation uses persisted tools”

**Suggestion:** even without big refactors, consider separating into clearer units over time:
- `DiskAgentRepository` (load/save)
- `AgentRuntimeConfigurer` (tools/context/mcp/critic)
- `EnvOverridePolicy` (validate + merge behavior)

---

## [IMPROVEMENT OPPORTUNITIES] (good taste / simpler invariants)

### 4) The env override “required vars” error message hardcodes model requirement
**File:** `openhands_cli/stores/agent_store.py` (`MissingEnvironmentVariablesError`)

Message says you must set `LLM_MODEL`. That’s a policy choice.

If you truly want the invariant “headless env override requires explicit model”, OK — but then don’t hint at defaults elsewhere. Otherwise, consider making `LLM_MODEL` optional with a default (better UX), and update the invariant + docs.

---

### 5) API client now loads only from disk; clarify API surface
**File:** `openhands_cli/auth/api_client.py`

You changed:
- `existing_agent = store.load()` → `store.load_from_disk()`

This is probably correct (you want to compare persisted config), but it highlights that `AgentStore` has two very different “load” concepts:
- persisted vs runtime-configured.

**Suggestion:** rename for clarity:
- `load_from_disk` → `load_persisted_agent`
- `load_or_create` → `load_runtime_agent` (or similar)

Names are invariants.

---

### 6) Tests needed monkeypatch signature updates — indicates config knowledge is spread
Many tests had to change monkeypatch lambdas for `SettingsScreen.is_initial_setup_required` because new parameters were threaded in.

This isn’t a functional bug, but it’s a smell: env/headless mode knowledge is being pushed deep into UI code, increasing surface area.

---

## [STYLE NOTES] (minor)

### 7) `LLMEnvOverrides.has_overrides()` uses `any([..])`
Prefer `return self.api_key is not None or ...` (or `any(x is not None for x in (...))`). Current code works, but it’s slightly sloppy and easy to extend incorrectly.

---

## VERDICT
✅ **Worth merging if** you tighten the policy docs and reduce the “mode flag plumbing” risk.  
❌ **Needs rework** if the project strongly cares about long-term invariants around configuration sources (disk vs env vs runtime config).

**Key insight:** you’re trending toward “configuration as implicit global mode flags passed everywhere”. The fix removed the *global variable* (good), but replaced it with *distributed parameters* (still risky).

malhotra5:
Will make docstring fixes 


I think I will merge this for now as it has critical bug fixes, but state management in the CLI is becoming a growing issue; improving it will be my next PR

enyst:
Interesting. I'm a bit sleepy and maybe I don't agree with all it said, but this is interesting:
> ❌ Needs rework if the project strongly cares about long-term invariants around configuration sources (disk vs env vs runtime config).

I think we do care. Not sure if the agent's solution is too good (or maybe I didn't fully understand it), and timing doesn't matter IMHO (we can merge now and rethink later) as much as getting this right, I say this because what I keep thinking of for a while and trying things, is how to make agents "get it" - because I think, as agents become long-running / multi-agent, code itself will matter less... but the code design constraints will matter more.

Imagine you leave the agents go wild and merge PRs. There can always be some weird code or mocking unit tests so that they test the library, not the stuff under test. 😅 But if there is docs on code design constraints, if they are embedded in the code design in a visible way, sooner or later a "reviewer" agent or "tech debt guard" will notice something's wrong.

Well, at least that's my thinking atm. 🤷 Happy for any thoughts!

malhotra5:
Yes I like the idea of constraints! 

I've been structuring my PR descriptions (e.g https://github.com/OpenHands/OpenHands-CLI/pull/277) with itemized requirements defining the north star for the intended behavior. But constraints seems like a natural next step to establish **HOW** we want to get there

Been talking with Xingyao about automating cycles between coding and codereview roasted between multiple agents, but I think having the user specify more concrete constraints as part of the process would be huge

PR #263, opened 2025-12-28
Always create fresh condenser with current defaults on load

xingyaoww:
## Summary

This PR ensures existing users get the latest condenser settings (e.g., max_size, keep_first) without needing to reconfigure. Previously, condenser settings were loaded from agent_settings.json which meant users would keep using old defaults even after SDK updates.

## Changes

- Modified `agent_store.py` to always create a fresh `LLMSummarizingCondenser` with current SDK defaults when loading agent settings, rather than preserving the old condenser configuration

## Testing

This change ensures users automatically benefit from improved condenser defaults in new SDK versions.

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@openhands/test-condensation-fix
```

PR #390, opened 2026-01-23
Preserve condenser settings from agent_settings.json

### LINKED ISSUE #389: agent_settings.json condenser settings overridden by defaults on load
### Problem
When a user already has `~/.openhands/agent_settings.json` on disk with custom condenser settings (e.g. `LLMSummarizingCondenser.max_size`, `keep_first`), loading the agent replaces the condenser with a new default condenser, losing user settings.

### Expected
User-provided condenser settings in `agent_settings.json` should be preserved. Only runtime-only changes (env LLM overrides + telemetry metadata) should be applied.

### Actual
`AgentStore.load()`/runtime config path creates `LLMSummarizingCondenser(llm=...)` without carrying over `max_size`/`keep_first`, effectively resetting to defaults.

### Repro
1. Create `~/.openhands/agent_settings.json` with `condenser.max_size=123` and `keep_first=7`.
2. Run CLI / call `AgentStore().load()`.
3. Observe loaded condenser has `max_size=240` and `keep_first=2` (defaults).

### Location
`openhands_cli/stores/agent_store.py` runtime config section that recreates condenser.

enyst:
HUMAN:

Fixes #389

@xingyaoww please correct me if wrong, my feeling is that we introduced this behavior in the middle of some condenser bug fixes, in particular one where we made the default `max_size=240` if I recall correctly. So we wanted users to use that, rather than 120 which ran into the bug.

This seemed to me an emergency, it doesn’t seem the right behavior long term though. My condenser is set to 700 or so… 😅

—-

AGENT:

### Summary
Stop recreating the condenser with default values during agent load. Instead, preserve the user-configured `LLMSummarizingCondenser` fields from `agent_settings.json` and only update its `llm` with:
- env var overrides (when enabled)
- runtime metadata (`litellm_extra_body`)

### Why
User settings on disk should not be overwritten by defaults at runtime.

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/cf942108741f4986b3b7b19e8205e4d3)

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@fix-condenser-defaults-not-overriding
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,17 @@ The first time you run the CLI, it will guide you through configuring your LLM s
 openhands
 ```
 
+
+### Configuration
+
+OpenHands CLI stores configuration under `~/.openhands/` (created on first run):
+
+- `agent_settings.json`: persisted agent settings (including condenser config)
+- `cli_config.json`: CLI/TUI preferences (e.g., critic enabled)
+- `mcp.json`: MCP server configuration
+
+By default, environment variables like `LLM_API_KEY`, `LLM_MODEL`, and `LLM_BASE_URL` are ignored; pass `--override-with-envs` to apply them (not persisted).
+
 ### Running Modes
 
 | Mode | Command | Best For |
@@ -125,7 +136,7 @@ openhands --headless --json -t "Create a Flask app"
 
 ## Documentation
 
-For complete documentation, visit [docs.openhands.dev/openhands/usage/cli]().
+For complete documentation, visit https://docs.openhands.dev/openhands/usage/cli.
 
 ## License
 
--- a/openhands_cli/stores/agent_store.py
+++ b/openhands_cli/stores/agent_store.py
@@ -1,4 +1,4 @@
-# openhands_cli/settings/store.py
+# openhands_cli/stores/agent_store.py
 from __future__ import annotations
 
 import json
@@ -458,10 +458,7 @@ class AgentStore:
 
         Args:
             llm_api_key: The LLM API key to use
-            settings: User settings dictionary containing model and other config
-            base_url: Base URL for the LLM service (defaults to
-                `settings['llm_base_url']`
-            )
+            settings: User settings dictionary (e.g., "llm_model", "llm_base_url")
             default_model: Default model to use if not specified in settings
 
         Returns:


malhotra5:
This behavior was introduced in #263

enyst:
> This behavior was introduced in #263

I updated the description to expand on why I think this happened, Xingyao is most welcome to correct me if wrong. It just … feels upside down IMHO. The user customizes their experience, I think maybe the SDK shouldn’t decide in their place 🤔

malhotra5:
Ah I see where you're coming from

@xingyaoww please feel to provide more context. If I recall correctly, we had some condenser settings which were causing nasty crashes in a running conversation, updating those defaults in addition to some other changes helped fix the problem?

As of today we don't expose a UI in the CLI to configure the condenser parameters themselves, which is why overriding with the default every time was considered 

I do think we want to think carefully about whether we want to expose the condenser parameters to an end user or not (maybe its mostly for researchers to easily tweak the params and run experiments, less for an end user to meddle with)

enyst:
> If I recall correctly, we had some condenser settings which were causing nasty crashes in a running conversation, updating those defaults in addition to some other changes helped fix the problem?

I think it comes from here:

https://github.com/OpenHands/software-agent-sdk/issues/1518#issuecomment-3695059492

The issue wasn’t caused by the settings, it started happening after some changes to the condenser algorithm. The error “0 events found to condense” happened. I think the quick fix - double the size - was an attempt to mitigate until we find a better fix. We guessed that if the condenser has more history to work with (240 instead of 120), it will find events maybe, instead of finding 0 events.

But finding 0 events was not about settings. Calvin already made other fixes and is working on improving it.

> As of today we don't expose a UI in the CLI to configure the condenser parameters themselves, which is why overriding with the default every time was considered

Ah, good point.

enyst:
Just to clarify, the condenser settings defaults haven’t changed for six months or more. They haven’t caused crashes, they were just low. For different reasons than potentially being a short term mitigation, I wanted to increase them because they’re too low for most or all SOTA LLMs.

They just say how many events to summarize, or what token limit should be for history. They’re something many people need to experiment with, or tweak for their LLM or task.

We are getting requests to make them configurable easily:
- https://github.com/OpenHands/OpenHands-CLI/issues/396

enyst:
I think we covered the tweak in main already, my agent wanted to point out some doc mismatches when we talked about this, though, so I'll just let them in.

PR #400, opened 2026-01-24
fix: display agent MessageEvents in TUI when critic is enabled

### LINKED ISSUE #399: [Bug] CLI TUI does not reliably display MessageEvents when critic is enabled
## Version
OpenHands CLI v1.10.0

## Summary
When the critic is enabled, the most recent agent events (particularly MessageEvents with critic_result) are not always being displayed in the terminal UI. The agent completes its work and events are properly registered/persisted to disk, but the terminal display appears frozen/stuck, misleading users into thinking the agent is hung.

This is an intermittent problem. I have seen two instances and and third conversation where the TUI appeared to work appropriately.

## Symptoms

Users observe the TUI appearing frozen after the agent completes a response:
- The agent thinks and the display stops
- Users ask "are you stuck?" or "What is happening?"
- Users report: "I am in the CLI and the last event or activity is not displaying"

### Evidence from conversation analysis

In affected conversations, there is a significant time gap between:
- When a MessageEvent with critic_result is registered (persisted to disk)
- When the user realizes nothing is displayed and sends a follow-up message

Example timeline:
```
15:42:43 - Agent sends MessageEvent with critic_result (score=0.78) - NOT DISPLAYED
15:58:01 - User asks "are you stuck?" (~15 minute gap)
15:58:08 - Agent responds "No, I'm not stuck! I actually completed the analysis..." - NOT DISPLAYED
```

### Affected Event Types
Events with `critic_result` field attached are not being streamed/displayed to the terminal:

```json
{
  "id": "<event-id>",
  "timestamp": "<timestamp>",
  "source": "agent",
  "kind": "MessageEvent",
  "llm_message": { ... },
  "critic_result": {
    "score": 0.7766118049621582,
    "message": "Success: 0.78; {...}",
    "metadata": { ... }
  }
}
```

### Key Observations
1. Events ARE properly persisted to the event log files on disk
2. Events ARE NOT displayed in the terminal/CLI output
3. The issue specifically affects events containing `critic_result`
4. ActionEvents and ObservationEvents (without critic) display normally
5. The agent continues working correctly - only the display is affected

## Suspected Root Cause
The CLI terminal renderer/event stream handler may not be properly handling events that include the `critic_result` field, causing them to be silently dropped from the display pipeline while still being persisted to storage.

Possible locations to investigate:
- Event streaming/rendering code in CLI
- MessageEvent handler when critic_result is present
- Event filter/transform logic that may be inadvertently dropping critic-enriched events

## Impact
- **Severity: Medium-High** - Users perceive the system as stuck/frozen
- **User Experience**: Confusing and frustrating; users don't know the agent has completed
- **Workaround**: None identified - users must wait or interrupt and check event logs manually

## Steps to Reproduce
1. Enable critic mode in OpenHands CLI v1.10.0
2. Start a conversation that triggers agent MessageEvents
3. Observe that when the agent sends a MessageEvent with critic evaluation, the terminal does not display it
4. Terminal appears frozen until user sends another message

## Expected Behavior
All agent events (including those with critic_result) should be streamed and displayed in the terminal in real-time.

## Actual Behavior
MessageEvents with critic_result are persisted to disk but not displayed in the terminal, making the UI appear frozen.

## Related PRs/Issues

- **PR #360** "Add Critic Result Display in TUI" (merged) - Added the critic display feature
- **PR #388** "(Hotfix): fix headless mode + disable critic in headless mode" (open) - Related fix for headless mode
- **PR #391** "Fix critic feedback text" (open) - Minor text fix for critic

## Environment
- OpenHands CLI v1.10.0
- macOS (also likely affects other platforms)

jpshackelford:
## Summary

Fixes #399 - CLI TUI does not display MessageEvents when critic is enabled

## Problem

Agent MessageEvents (including those with `critic_result`) were silently dropped in the TUI due to a bug in the delegation filtering logic in `richlog_visualizer.py`. The condition:

```python
if not (event.sender and self._name):
    return None
```

Would filter out ALL MessageEvents without a sender, even when the visualizer has a name set (which is the normal case for the main conversation).

## Reproduction Case

**Steps to reproduce:**

```bash
$ openhands
```

Then ask a simple conversational question (one that doesn't require tool use):
```
What is the capital of France?
```

**Without the fix:** The agent's response is NOT displayed in the TUI. The UI appears stuck after "Thinking...".

**With the fix:** The response displays normally.

**Evidence in conversation logs:**

Check `~/.openhands/conversations/<latest>/events/` - you'll see three events:

```
event-00000: SystemPromptEvent
event-00001: MessageEvent (user question)
event-00002: MessageEvent (agent response) ← This one is dropped!
```

The agent's response event shows:
```
kind: MessageEvent
source: agent
role: assistant
has_critic_result: True
sender: None           ← No sender because this is NOT a delegation message
content: "The capital of France is **Paris**."
```

The bug filters out this event because `sender` is `None`, even though it's a valid agent response that should be displayed.

## Solution

Restructured the MessageEvent handling logic into three clear cases:

1. **Delegation messages** (sender AND name set) - format with arrow notation
2. **Regular agent messages** (name set, no sender, assistant role) - display as markdown ← **NEW**
3. **No name context** - skip (not used in CLI)

User messages without a sender are still skipped as they are displayed separately in the UI.

## Testing

Added new test class `TestAgentMessageEventDisplay` with 4 tests:
- `test_agent_message_event_displays_when_name_set_no_sender` - primary bug reproduction
- `test_agent_message_event_with_critic_result_displays` - critic-specific case
- `test_user_message_still_skipped_when_skip_user_messages_true` - regression test
- `test_delegation_message_still_works_with_sender` - regression test

All 48 tests in `test_richlog_visualizer.py` pass.

## Root Cause Analysis

Full analysis is documented in the issue comments:
- [Root cause identification](https://github.com/OpenHands/OpenHands-CLI/issues/399#issuecomment-3795611706)
- [OpenHands investigation notes](https://github.com/OpenHands/OpenHands-CLI/issues/399#issuecomment-3795625405)
- [Explanation of intermittent behavior](https://github.com/OpenHands/OpenHands-CLI/issues/399#issuecomment-3795634926)

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@fix/message-event-display-bug
```

### DIFF
--- a/tests/tui/widgets/test_richlog_visualizer.py
+++ b/tests/tui/widgets/test_richlog_visualizer.py
@@ -979,7 +979,6 @@ class TestAgentMessageEventDisplay:
         from textual.widgets import Markdown
 
         from openhands.sdk import Message
-
         from openhands_cli.tui.widgets.collapsible import Collapsible
 
         app = App()
@@ -1007,7 +1006,7 @@ class TestAgentMessageEventDisplay:
             "Agent MessageEvent without sender should still display when "
             "visualizer has name set. This is the normal case for CLI usage."
         )
-        assert isinstance(widget, (Markdown, Collapsible))
+        assert isinstance(widget, Markdown | Collapsible)
 
     def test_agent_message_event_with_critic_result_displays(self):
         """Agent MessageEvent with critic_result should display.
@@ -1032,7 +1031,7 @@ class TestAgentMessageEventDisplay:
 
         message = Message(
             role="assistant",
-            content=[TextContent(text="Based on my analysis, here are the findings...")],
+            content=[TextContent(text="Based on my analysis, here are the findings")],
         )
         # Create event with critic_result (simulating what the SDK does)
         event = MessageEvent(llm_message=message, source="agent")


malhotra5:
Hmm the visualizer with agent delegation is not released yet; was the missing message events something you're seeing in the latest released CLI?

jpshackelford:
That's a great question. I do sometimes run from main. But in this case I had upgraded to 1.10.0 using `$ uv tool upgrade openhands`.  I have reverified that the reproduction steps above do reproduce the issue with 1.10.0.

jpshackelford:
Correction. If I install with `uv tool install openhands==1.10.0 --force` I cannot reproduce. So it seems that when I thought I was upgrading to 1.10.0 I was upgrading to a newer version of main. 

That said--this is still a bug on main and this is still a fix on main. It was just incorrectly reported against the 1.10.0 release.

PR #421, opened 2026-01-26
refactor: add reactive state management + decouple main app from conversation management

malhotra5:
## Problem

The main app was a god object handling too many things. It handles system commands, user prompts, screen coordinations, conversation lifecycle management, and more. 

Furthermore, the app would manually update UI states (e.g updating the splash banner when a conversation is created). We refactor this so that UI states can self update based on reactive states. We separate conversation state and lifecycle responsibilities and more. OpenHands has left nice comments regarding the differences between the original architecture and the one we're aiming for. 

The goal of this refactor is rework state management, but retain all of the existing functionality (we build confidence in this refactor via existing unit tests and most importantly the E2E snapshot tests)

In future PR, we will make additional changes such as 

- make confirmation panel mount based on messages rather than a blocking callback 
- make sure we can mount screens without direct references to openhands app from child components to further decouple (follow the pattern for `HistoryPanel` instead the one for `InlineConfirmationPanel`) 
- etc

NOTE: we use the current hierarchy of widgets because 

1. Messages only propagate to parents in the DOM. To send information to a sibling it needs a direct reference or needs to be passed via a shared parent node
2. Data binding only works when composing the binding elements from directly within the class. Hence why `ConversationState` needs to be a textual component of its own, rather than a standalone object 

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@refactor/improve-state-management-v2
```

### DIFF
--- a/tests/snapshots/e2e/conftest.py
+++ b/tests/snapshots/e2e/conftest.py
@@ -20,6 +20,7 @@ from tests.conftest import create_test_agent_config
 from tui_e2e.mock_llm_server import MockLLMServer
 from tui_e2e.trajectory import get_trajectories_dir, load_trajectory
 
+
 if TYPE_CHECKING:
     from syrupy.types import SerializableData, SerializedData
 
@@ -60,7 +61,9 @@ def normalize_svg_colors(svg: str) -> str:
 
     # Sort colors to create deterministic mapping
     # Group by (color, extra_styles) to handle font-weight etc.
-    sorted_colors = sorted(old_classes.items(), key=lambda x: (x[1][0], x[1][1], int(x[0])))
+    sorted_colors = sorted(
+        old_classes.items(), key=lambda x: (x[1][0], x[1][1], int(x[0]))
+    )
 
     # Create mapping from old class number to new class number
     class_mapping = {}
@@ -70,13 +73,12 @@ def normalize_svg_colors(svg: str) -> str:
         new_class_defs.append(f".terminal-r{new_num} {{ fill: {color}{extra} }}")
 
     # Remove old class definitions and insert new sorted ones
-    # Find the block of class definitions
     def remove_class_defs(match):
         return ""
 
     svg = re.sub(color_pattern, remove_class_defs, svg)
 
-    # Find where to insert the new class definitions (after the last @font-face or at the start of style)
+    # Find where to insert the new class definitions
     style_match = re.search(r"(<style[^>]*>.*?)(</style>)", svg, re.DOTALL)
     if style_match:
         style_content = style_match.group(1)
@@ -84,7 +86,9 @@ def normalize_svg_colors(svg: str) -> str:
 
         # Insert the new class definitions before </style>
         new_class_block = "\n    " + "\n".join(new_class_defs) + "\n    "
-        svg = svg.replace(style_match.group(0), style_content + new_class_block + style_end)
+        svg = svg.replace(
+            style_match.group(0), style_content + new_class_block + style_end
+        )
 
     # Replace class references in the SVG content
     # Match class="terminal-rN" or class="terminal-XXXXX-rN"
@@ -113,13 +117,16 @@ class ColorNormalizedSVGExtension(SingleFileSnapshotExtension):
     ) -> Optional["SerializableData"]:
         """Normalize SVG data right after they are loaded from persistent storage."""
         data = super()._read_snapshot_data_from_location(*args, **kwargs)
-        if data is not None:
+        if data is not None and isinstance(data, str):
             data = normalize_svg_colors(data)
         return data
 
     def serialize(self, *args, **kwargs) -> "SerializedData":
         """Normalize SVG data before comparison and persistence."""
-        return normalize_svg_colors(super().serialize(*args, **kwargs))
+        data = super().serialize(*args, **kwargs)
+        if isinstance(data, str):
+            return normalize_svg_colors(data)
+        return data
 
 
 @pytest.fixture


openhands_cli/tui/textual_app.py
```diff
@@ -15,11 +15,10 @@
```
malhotra5, line 43: @openhands please look at the method to handle commands and all the methods that handle them individually

please move the logic out of this class and to another, and pass the command handling logic to the other class instead

you can pass over the app object itself if its required, but do the command handling elsewhere

malhotra5:
@openhands /codereview-roasted. think carefully about separation of concerns and overall maintainability for state management. post back the review as a github comment

malhotra5:
@openhands /codereview-roasted. think carefully about separation of concerns and overall maintainability for state management. post back the review as a github comment

malhotra5:
## 🔴 **Needs Improvement** - Violates Fundamental Principles

This PR attempts to solve a real problem (scattered state management) but creates a bigger one: a **God Object** masquerading as a "centralized state manager."

---

**[openhands_cli/tui/core/state.py, Lines 1-572] 🔨 Data Structure: ConversationView is a God Object**

`ConversationView` is simultaneously:
- A `Container` widget (UI composition)
- A state manager (reactive vars)
- A conversation lifecycle manager (attach_conversation, reset_conversation_state)
- A ConversationRunner factory (get_or_create_runner)
- A message handler (ProcessUserInput, NewConversationRequested)
- A policy synchronizer (_sync_policy_to_conversation)

This violates the Single Responsibility Principle so hard it should be a felony. The name "ConversationView" suggests it's a *view*, but it's actually a view + model + controller + factory all rolled into one 572-line monstrosity.

**Suggestion**: Separate concerns:
1. `ConversationState` - Pure state container with reactive vars (no UI, no runner)
2. `ConversationView` - UI composition only, binds to state
3. `ConversationLifecycle` - Runner creation, conversation attachment

---

**[openhands_cli/tui/textual_app.py, Lines 148-170] 🔨 Redundant Aliases Create Confusion**

```python
self.conversation_view = ConversationView(...)
# Backwards compatibility alias
self.app_state = self.conversation_view
self.state_manager = self.conversation_view
```

Three names for the same object is not "backwards compatibility" - it's confusion. Pick ONE name and use it consistently. The codebase now has:
- `app.conversation_view`
- `app.app_state`
- `app.state_manager`

All pointing to the same thing. This is the opposite of clarity.

---

**[openhands_cli/tui/messages.py, Lines 54-69] 🔨 Unnecessary Message Indirection**

```python
class ProcessUserInput(Message):
    """Note: This is now largely superseded by UserInputSubmitted..."""
```

If it's "largely superseded," why does it exist? The message flow is:
```
InputField → UserInputSubmitted → InputAreaContainer → ProcessUserInput → ConversationView
```

Why two messages? `UserInputSubmitted` could bubble directly to `ConversationView`. The intermediate `ProcessUserInput` adds complexity without value.

---

**[openhands_cli/tui/core/state.py, Lines 243-260] 🔨 Thread Safety is Fragile**

```python
def __init__(self, ...):
    self._main_thread_id = threading.current_thread().ident
    ...

def on_mount(self) -> None:
    self._main_thread_id = threading.current_thread().ident  # Set again?
```

`_main_thread_id` is set in `__init__` AND `on_mount`. What if they're different threads? This is a race condition waiting to happen. Textual has `self.app.call_from_thread()` - use it unconditionally instead of this fragile thread detection.

---

**[openhands_cli/tui/widgets/input_area.py, Lines 1-228] 📍 InputAreaContainer Does Too Much**

This "container" handles:
- Rendering user messages to a *sibling* widget (scroll_view)
- Executing slash commands
- Clearing content on conversation_id change
- Accessing `self.app` for modals, notifications, history

A container that reaches into its sibling to mount widgets is a code smell. The rendering logic should be in `ConversationView` or `ScrollableContent`, not in `InputAreaContainer`.

---

**[openhands_cli/tui/widgets/status_line.py, Lines 146-190] 📍 View Doing Model Work**

```python
def _get_metrics_display(self) -> str:
    # Extract values from metrics object
    if self.metrics is None:
        return "ctx N/A • $ 0.00 (↑ 0 ↓ 0 cache N/A)"

    # Calculate cache hit rate
    if input_tokens > 0:
        cache_hit_rate = f"{(cache_read_tokens / input_tokens * 100):.0f}%"
```

The view is calculating cache hit rates. This logic belongs in the `Metrics` object or a helper function, not in the status line widget.

---

**[openhands_cli/tui/widgets/splash.py, Lines 1-156] 📍 Dual Initialization is Confusing**

```python
# Reactive binding for conversation switching
yield SplashContent(id="splash_content").data_bind(
    conversation_id=ConversationView.conversation_id,
)

# Later, in OpenHandsApp:
splash_content.initialize(has_critic=has_critic)
```

Two initialization mechanisms: reactive binding AND imperative `initialize()`. This is confusing. Either make `has_critic` a reactive property, or don't use reactive binding at all.

---

**[openhands_cli/tui/core/conversation_runner.py, Lines 80-85] 📍 State Duplication**

```python
self._running = False
...
def _update_run_status(self, is_running: bool) -> None:
    self._running = is_running
    self._conversation_view.set_running(is_running)
```

`ConversationRunner._running` and `ConversationView.running` both track the same state. This is a recipe for desynchronization bugs. The runner should query `ConversationView.running` instead of maintaining its own copy.

---

**[openhands_cli/tui/core/state.py, Lines 26-28] 📝 Misleading Module Docstring**

```python
"""Centralized state management for OpenHands TUI.

This module provides ConversationView, a centralized state container that:
```

The docstring says "state container" but the class is a `Container` widget. This naming confusion permeates the entire design.

---

### **VERDICT:**
❌ **Needs rework**: The fundamental design conflates UI composition with state management. Splitting `ConversationView` into separate state and view components would make this maintainable.

### **KEY INSIGHT:**
The PR correctly identifies that scattered state is a problem, but the solution creates a God Object that's harder to reason about than the original scattered state. Good state management should be *invisible* to the UI layer, not embedded in a widget.

---

*"Bad programmers worry about the code. Good programmers worry about data structures and their relationships."* - The data structure here (ConversationView) has too many relationships.

malhotra5:
proposal


```mermaid
flowchart TB
    subgraph UI["UI Components"]
        InputField
        ScrollableContent
        HistorySidePanel
        StatusLines
        SplashContent
    end

    subgraph STATE["State Layer"]
        ConversationState["ConversationState<br/>(reactive properties only)"]
    end

    subgraph MANAGER["ConversationManager"]
        direction TB
        CREATE["Create Conversation"]
        SWITCH["Switch Conversation"]
        SEND["Send Message"]
        RUNNERS["Runner Registry"]
    end

    subgraph SERVICES["Services"]
        InputAreaContainer
        ConversationRunner
    end

    subgraph EXTERNAL["External"]
        SDK
        LocalFileStore
    end

    %% UI to Manager (direct messaging)
    InputField -->|"SendMessage"| MANAGER
    InputAreaContainer -->|"CreateConversation"| MANAGER
    InputAreaContainer -->|"SwitchConversation"| MANAGER
    HistorySidePanel -->|"SwitchConversation"| MANAGER

    %% Manager operations
    MANAGER -->|"creates/manages"| ConversationRunner
    MANAGER --> LocalFileStore
    ConversationRunner <--> SDK

    %% Manager updates state
    MANAGER -->|"updates"| ConversationState

    %% UI binds to state (reactive)
    ConversationState <-.->|"data_bind"| ScrollableContent
    ConversationState <-.->|"data_bind"| StatusLines
    ConversationState <-.->|"data_bind"| SplashContent
    ConversationState <-.->|"watch"| HistorySidePanel

    %% Runner events to UI
    ConversationRunner -->|"events"| ScrollableContent

    classDef ui fill:#e8f5e9,stroke:#2e7d32
    classDef state fill:#e3f2fd,stroke:#1565c0
    classDef manager fill:#fff3e0,stroke:#ef6c00,stroke-width:2px
    classDef service fill:#f3e5f5,stroke:#7b1fa2
    classDef external fill:#fafafa,stroke:#757575

    class InputField,ScrollableContent,HistorySidePanel,StatusLines,SplashContent ui
    class ConversationState state
    class MANAGER,CREATE,SWITCH,SEND,RUNNERS manager
    class InputAreaContainer,ConversationRunner service
    class SDK,LocalFileStore external
```



Proposed architecture

malhotra5:
## Updated Architecture Diagram

This PR implements the following architecture:

### System Overview

```mermaid
flowchart TB
    subgraph UI["UI Components"]
        InputField
        ScrollableContent
        HistorySidePanel
        StatusLines
        SplashContent
    end

    subgraph STATE["State Layer"]
        ConversationState["ConversationState<br/>(reactive properties,<br/>metrics via attached conversation)"]
    end

    subgraph MANAGER["ConversationManager (Container)"]
        direction TB
        CREATE["CreateConversation"]
        SWITCH["SwitchConversation"]
        SEND["UserInputSubmitted"]
        POLICY["SetConfirmationPolicy"]
        RUNNERS["Runner Registry"]
    end

    subgraph SERVICES["Services"]
        InputAreaContainer
        ConversationRunner
    end

    subgraph EXTERNAL["External"]
        SDK["BaseConversation"]
        LocalFileStore
    end

    InputField -->|"UserInputSubmitted<br/>(bubbles)"| MANAGER
    InputAreaContainer -->|"CreateConversation<br/>(bubbles)"| MANAGER
    InputAreaContainer -->|"SetConfirmationPolicy<br/>(bubbles)"| MANAGER
    HistorySidePanel -->|"SwitchConversation<br/>(bubbles)"| MANAGER
    ConversationRunner -->|"SetConfirmationPolicy<br/>(via MessagePump)"| MANAGER

    MANAGER -->|"creates/manages"| ConversationRunner
    MANAGER -->|"policy sync"| SDK
    MANAGER --> LocalFileStore
    ConversationRunner <--> SDK

    MANAGER -->|"updates"| ConversationState
    ConversationState <-.->|"attach_conversation<br/>(for metrics)"| SDK

    ConversationState <-.->|"data_bind"| ScrollableContent
    ConversationState <-.->|"data_bind"| StatusLines
    ConversationState <-.->|"data_bind"| SplashContent
    ConversationState <-.->|"watch()"| HistorySidePanel

    ConversationRunner -->|"events"| ScrollableContent
```

### Widget Hierarchy

```mermaid
graph TB
    subgraph APP["OpenHandsApp"]
        subgraph MGR["ConversationManager (Container)"]
            NOTE["Messages bubble here"]
            subgraph CONTENT["Horizontal(#content_area)"]
                subgraph STATE["ConversationState"]
                    SCROLL["ScrollableContent"]
                    INPUT["InputAreaContainer"]
                end
                HSP["HistorySidePanel<br/>(dynamic)"]
            end
        end
    end

    MGR -.->|"owns"| RUNNER["ConversationRunner"]
    RUNNER -.->|"manages"| CONV["BaseConversation"]
    STATE -.->|"attach_conversation"| CONV
```

### Key Design Decisions

| Component | Responsibility |
|-----------|----------------|
| **ConversationManager** | Container that wraps content. Handles all operations. Owns runners. Syncs policy to conversation. |
| **ConversationState** | Pure reactive state for UI binding. Reads metrics from attached conversation. |
| **ConversationRunner** | Uses `MessagePump` interface to post policy changes back to Manager. |

### Message Flow Pattern

- **Bubbling**: Messages from children (`InputField`, `InputAreaContainer`, `HistorySidePanel`) bubble up to `ConversationManager`
- **MessagePump**: `ConversationRunner` posts `SetConfirmationPolicy` via the `MessagePump` interface
- **Components**: Just use `self.post_message()` - no explicit targeting needed

### Separation of Concerns

| Concern | Owner |
|---------|-------|
| Policy sync to conversation | ConversationManager |
| Metrics reading from conversation | ConversationState (via timer) |
| Runner lifecycle | ConversationManager |
| Reactive state for UI | ConversationState |

malhotra5:
## Current Architecture (main branch)

This shows the state management architecture **before** the refactoring in this PR.

### System Overview

```mermaid
flowchart TB
    subgraph USER["👤 User"]
        E1["Types message"]
        E2["Types slash command"]
        E3["Opens history & switches"]
    end

    subgraph APP["OpenHandsApp (God Object)"]
        direction TB
        CONV_ID["conversation_id"]
        CONV_RUNNER["conversation_runner"]
        SIGNAL["conversation_running_signal"]
        HANDLERS["@on() Message Handlers"]
        CMD_HANDLER["_handle_command()"]
        MSG_HANDLER["_handle_user_message()"]
        CREATE_RUNNER["create_conversation_runner()"]
    end

    subgraph SERVICES["Services"]
        CM["ConversationManager"]
        CS["ConversationSwitcher"]
        CR["ConversationRunner"]
    end

    subgraph UI["UI Components"]
        InputField
        HistorySidePanel
        MainDisplay["main_display"]
        StatusLines
    end

    subgraph EXTERNAL["External"]
        SDK["BaseConversation"]
        LocalFileStore
    end

    E1 --> InputField
    E2 --> InputField
    E3 --> HistorySidePanel

    InputField -->|"InputField.Submitted"| APP
    APP --> CMD_HANDLER
    APP --> MSG_HANDLER

    CMD_HANDLER -->|"/new"| CM
    CMD_HANDLER -->|"/history"| HistorySidePanel
    MSG_HANDLER --> CREATE_RUNNER
    MSG_HANDLER --> CR

    HistorySidePanel -->|"SwitchConversationRequest"| APP
    APP -->|"forwards"| CM
    CM --> CS
    CS -->|"directly modifies"| APP

    CM -->|"posts"| ConversationCreated
    CM -->|"posts"| ConversationTitleUpdated
    CS -->|"posts"| ConversationSwitched

    ConversationCreated -->|"forwarded"| HistorySidePanel
    ConversationTitleUpdated -->|"forwarded"| HistorySidePanel
    ConversationSwitched -->|"forwarded"| HistorySidePanel

    CR -->|"callback"| SIGNAL
    SIGNAL -->|"subscribe"| StatusLines
    CR <--> SDK
    CM --> LocalFileStore

    classDef god fill:#ffcdd2,stroke:#c62828,stroke-width:2px
    classDef service fill:#fff3e0,stroke:#ef6c00
    classDef ui fill:#e8f5e9,stroke:#2e7d32
    classDef external fill:#fafafa,stroke:#757575

    class APP,CONV_ID,CONV_RUNNER,SIGNAL,HANDLERS,CMD_HANDLER,MSG_HANDLER,CREATE_RUNNER god
    class CM,CS,CR service
    class InputField,HistorySidePanel,MainDisplay,StatusLines ui
    class SDK,LocalFileStore external
```

### Widget Hierarchy

```mermaid
graph TB
    subgraph APP["OpenHandsApp"]
        subgraph CONTENT["Horizontal(#content_area)"]
            subgraph MAIN["VerticalScroll(#main_display)"]
                SPLASH["Splash widgets"]
                DYNAMIC["Dynamic widgets"]
            end
            subgraph INPUT_AREA["Container(#input_area)"]
                WSL["WorkingStatusLine"]
                IF["InputField"]
                ISL["InfoStatusLine"]
            end
            HSP["HistorySidePanel<br/>(dynamic)"]
        end
        FOOTER["Footer"]
    end

    APP -.->|"owns"| CR["ConversationRunner"]
    APP -.->|"owns"| CM["ConversationManager"]
    CM -.->|"delegates to"| CS["ConversationSwitcher"]

    style APP fill:#ffcdd2,stroke:#c62828
```

### Message Flow Pattern

```mermaid
sequenceDiagram
    participant User
    participant InputField
    participant App as OpenHandsApp
    participant CM as ConversationManager
    participant CS as ConversationSwitcher
    participant HSP as HistorySidePanel
    participant CR as ConversationRunner

    Note over App: OpenHandsApp handles everything

    User->>InputField: Types message
    InputField->>App: InputField.Submitted
    App->>App: _handle_user_message()
    App->>CR: create_conversation_runner()
    App->>CR: process_message_async()
    CR-->>App: callback(running=true)
    App-->>App: Signal.publish()

    Note over App,HSP: Manual message forwarding required

    User->>InputField: /new
    InputField->>App: InputField.Submitted
    App->>App: _handle_command("/new")
    App->>CM: create_new()
    CM->>CM: directly modifies app.conversation_id
    CM->>App: post_message(ConversationCreated)
    App->>HSP: message forwarded by App

    User->>HSP: Click conversation
    HSP->>App: SwitchConversationRequest
    App->>CM: switch_to()
    CM->>CS: switch_to()
    CS->>App: directly modifies app.conversation_id
    CS->>App: post_message(ConversationSwitched)
    App->>HSP: message forwarded
```

### Problems with Current Architecture

| Issue | Description |
|-------|-------------|
| **God Object** | `OpenHandsApp` handles state, UI, commands, message routing |
| **Manual Message Forwarding** | App must manually route messages to HistorySidePanel |
| **Direct State Mutation** | `ConversationManager` and `ConversationSwitcher` directly modify `app.conversation_id` |
| **Callback-Based State** | `conversation_running_signal` uses Signal/callback pattern instead of reactive state |
| **No Reactive Binding** | Status lines subscribe to signals manually, no `data_bind()` |
| **Scattered Policy Logic** | Policy changes happen inside `ConversationRunner._change_confirmation_policy()` |
| **Tight Coupling** | Services hold reference to `OpenHandsApp` and modify it directly |

### Component Responsibilities (Current)

| Component | Responsibility |
|-----------|----------------|
| **OpenHandsApp** | Everything: state, UI, commands, message handling, runner creation |
| **ConversationManager** | Creates conversations, delegates switching to Switcher, posts messages |
| **ConversationSwitcher** | Handles switch logic, directly modifies App state |
| **ConversationRunner** | Runs agent, manages policy internally, uses callbacks for state updates |
| **HistorySidePanel** | Displays history, receives forwarded messages from App |
| **InputField** | Posts `Submitted` message to App |

### Message Types (Current)

| Message | Posted By | Handled By | Notes |
|---------|-----------|------------|-------|
| `InputField.Submitted` | InputField | OpenHandsApp | Direct handler |
| `ConversationCreated` | ConversationManager | App → HistorySidePanel | Forwarded |
| `ConversationSwitched` | ConversationSwitcher | App → HistorySidePanel | Forwarded |
| `ConversationTitleUpdated` | ConversationManager | App → HistorySidePanel | Forwarded |
| `SwitchConversationRequest` | HistorySidePanel | OpenHandsApp | App calls CM |
| `RevertSelectionRequest` | ConversationSwitcher | App → HistorySidePanel | Forwarded |

malhotra5:
I'm hoping to get the review process started, but will also look into cleaning up all the excessive doc strings etc

xingyaoww:
@OpenHands /codereview-roasted help me understand the changes in this pr and tell me which file/functions Re the most important that i should focus on reviewing ?

enyst:
@openhands Do a /codereview-roasted on this PR. Make sure you have `gh` installed so you can use it. Understand the proposed architecture, and understand the one from main branch, so you can focus on the important higher level changes, components and flow going from where to where. Then tell WDYT?

Respond in a comment on this PR, on github. Note that it will be rendered as markdown. Use smolpaws.

smolpaws:
🟡 **Taste Rating: Acceptable (but your god object just moved into a bigger condo).**

I get what you’re trying to do: split the OpenHandsApp blob into a ConversationManager + ConversationState + reactive UI. That’s the right direction. But you didn’t *actually* separate concerns — you repackaged them. The app is thinner, yes, but the new manager/state are still carrying UI scaffolding and lifecycle glue in ways that will keep biting you.

Here’s the roasted breakdown, focusing on **flow + architecture** (main vs PR):

**Main branch:**
- OpenHandsApp handled everything (state, switching, runner lifecycle, message routing).
- UI updates were imperative + forwarded messages.
- ConversationSwitcher mutated app state directly.

**This PR:**
- ConversationState owns reactive UI + *also* composes UI widgets.
- ConversationManager handles switching, runner registry, state updates, UI lifecycle bits.
- HistorySidePanel now watches state instead of forwarded messages.

It’s closer to correct, but the boundaries are still leaky.

---

## [CRITICAL ISSUES]

- **[tui/core/state.py:134–179] Data Structure / Ownership**
  ConversationState is supposed to be “pure reactive state” yet it *constructs UI widgets*. That forces UI composition into the data layer. This is not state; it’s a UI container wearing a state hat. You traded one god object for a hybrid monster.
  **Fix:** Let ConversationState be a plain reactive object; have a separate view/container compose widgets and bind to it. If Textual requires the reactive owner as message pump, wrap a View container that *owns* state and passes it down.

- **[tui/core/conversation_manager.py:192–519] Separation of Concerns**
  ConversationManager does: message dispatch, state mutation, UI focus control, modal orchestration, notifications, runner registry, and direct visualizer creation. That is still too much. It’s not “manager,” it’s “app-lite.”
  **Fix:** split “UI orchestration” (notifications, focus, modals) into App or a UI controller, and keep ConversationManager strictly about conversation lifecycle + runner coordination.

- **[tui/core/conversation_manager.py:320–403] Switch Flow Entanglement**
  Switching logic is now in ConversationManager *and* mutates UI state directly (notifications, input disable, scroll). That makes it impossible to unit-test switching without UI present.
  **Fix:** push UI transitions into app-layer callbacks or events; keep ConversationManager side-effect-free beyond updating ConversationState.

---

## [IMPROVEMENT OPPORTUNITIES]

- **[tui/panels/history_side_panel.py:186–252] Watchers are doing UI policy**
  The panel is now watching state for `is_switching` and inferring cancellation. That’s a cute hack, but state shouldn’t carry “UI cancellation recovery” semantics. It’s UI-specific state leaking back upward.
  **Fix:** introduce an explicit “switch_result” or “switch_cancelled” signal/event. Don’t overfit semantics into booleans.

- **[tui/core/state.py:287–327] Thread-safe mutation helpers**
  You added `set_running`, `set_metrics`, etc., but then you still write directly to `state.conversation_id` in some flows. That makes thread-safety a best-effort suggestion, not a contract.
  **Fix:** either enforce all state updates through the thread-safe API or drop the wrappers entirely. Half-measures rot.

- **[tui/core/conversation_manager.py:200–208] Title update on every message**
  State title is set on *every* user message, not just the first. That rewrites history titles on each input. If you intended “first message is title,” this is wrong (and it used to only update once).
  **Fix:** set title only if it’s currently empty.

---

## [STYLE NOTES]

- **[tui/messages.py] Message docstrings are now misleading**
  Some docs say messages bubble to ConversationState, but they actually bubble to ConversationManager. This is noise and misleads reviewers.

---

**VERDICT:** ❌ **Needs rework before merge.**

You’ve *reduced* the god object, but you’ve *not eliminated it.* The ConversationManager now owns both domain logic and UI choreography, and ConversationState is a UI container pretending to be pure state. This will make future refactors worse, not better.

**KEY INSIGHT:** You moved logic across files, but you didn’t change the *ownership model*. State, UI, and lifecycle are still tangled — just distributed.

malhotra5:
> [tui/core/state.py:134–179] Data Structure / Ownership
ConversationState is supposed to be “pure reactive state” yet it constructs UI widgets. That forces UI composition into the data layer. This is not state; it’s a UI container wearing a state hat. You traded one god object for a hybrid monster.
Fix: Let ConversationState be a plain reactive object; have a separate view/container compose widgets and bind to it. If Textual requires the reactive owner as message pump, wrap a View container that owns state and passes it down.


I looked into this in the past, but AFAIK its a textual limitation (you have to compose the widget you want to bind to). Will double check

> [tui/core/conversation_manager.py:192–519] Separation of Concerns
ConversationManager does: message dispatch, state mutation, UI focus control, modal orchestration, notifications, runner registry, and direct visualizer creation. That is still too much. It’s not “manager,” it’s “app-lite.”
Fix: split “UI orchestration” (notifications, focus, modals) into App or a UI controller, and keep ConversationManager strictly about conversation lifecycle + runner coordination.

> [tui/core/conversation_manager.py:320–403] Switch Flow Entanglement
Switching logic is now in ConversationManager and mutates UI state directly (notifications, input disable, scroll). That makes it impossible to unit-test switching without UI present.
Fix: push UI transitions into app-layer callbacks or events; keep ConversationManager side-effect-free beyond updating ConversationState.


Yep this is true, although I think this is largely due to conversation switching functionality and I felt reworking that would be out of scope for this refactor. Although happy to do this now as well, depending on whether you'd feel comfortable reviewing. Reason I say this is because I may simplify the user interactions as well (like not being able to switch in the middle of an ongoing conversation), which would therein simplify the code as well

pashkash:
> But you didn’t actually separate concerns — you repackaged them.

true; that was my goal: do not feed the good object but use small gods instead with the same dev style (as i tried to guess, cuz i am novice here)


@malhotra5 @enyst we can go further beyond god-managers and try to build redux way that will be much more easier to test (reducers (pure) + middlewares(side, io) = logic); use pydantic state; otherwise we will have e.g. toad in the end which doesn't make sense (its the same rich+textural) probably;
to consider:
UI Event → Action → Reducer → one Store → View
(controller here = dispatcher for actions)

Example: "Sending a Message"
UI (InputField): User types "Hello" and hits Enter.
-> Creates ActionSendMessage("Hello").
-> Calls store.dispatch(action).
Store:
-> Passes (State, Action) to the Reducer.
Reducer:
-> Creates a new State where "Hello" is added to the message list (status: "sending").
-> Returns the new State to the Store.
Store:
-> Saves the new State.
-> Notifies the UI.
-> Passes the Action to Middleware.
UI (ChatPane):
-> Receives notification. Re-renders. User sees their "Hello" message.
Middleware:
-> Sees ActionSendMessage.
-> "Aha, need to send to LLM." Calls runner.send("Hello").
-> ...time passes...
-> Receives response "How are you?".
-> Creates ActionAgentResponse("How are you?").
-> Calls store.dispatch(action).
Store -> Reducer -> UI:
Cycle repeats. Reducer adds the agent's response to the state. UI re-renders and shows the response.

Also I assume OH CLI will control autonomous fleets (plural), so it will be smth:
    state = {
      "agents": {
        "agent_1": { "status": "thinking", "task": "Fix bug" },
        "agent_2": { "status": "idle", "task": None },
        "agent_fleet_1": { "status": "running", "sub_agents": ["agent_1, agent_2"] }
      }
    }
with communications between agents/fleets via actions; (have no idea what it will be and i guess it will be inside sdk but still better to have simple extendable mechanics here); 
also we can work with cloud/local/whatever with just different separate middlewares



```mermaid
flowchart TB
    subgraph UI ["View Layer (Textual Widgets)"]
        direction TB
        Input[InputField]
        History[HistoryPanel]
        Chat[ChatPane]
        Layout[MainLayoutController]
    end

    subgraph CORE ["Core Logic (Pure Python)"]
        direction TB
        Action[Actions / Intents]
        Store[State Store]
        Reducer[Reducer / State Mutator]
    end

    subgraph ASYNC ["Async Middleware"]
        Runner[ConversationRunner]
        IO[File/Network IO]
        EffectManager[Effect Manager]
    end

    %% Data Flow
    Input -- "1. Textual Message" --> Layout
    History -- "1. Textual Message" --> Layout
    
    Layout -- "2. Dispatch(Action)" --> Store
    
    Store -- "3. Action" --> Reducer
    Reducer -- "4. New State" --> Store
    
    Store -- "5. Side Effect?" --> EffectManager
    EffectManager -- "6. Run Task" --> Runner
    Runner -- "7. Dispatch(ResultAction)" --> Store
    
    Store -.->|8. Signal/Callback| Layout
    Layout -.->|9. Re-render| Input & History & Chat

    classDef ui fill:#e8f5e9,stroke:#2e7d32
    classDef core fill:#e3f2fd,stroke:#1565c0
    classDef async fill:#fff3e0,stroke:#ef6c00
    
    class Input,History,Chat,Layout ui
    class Action,Store,Reducer core
    class Runner,IO,EffectManager async
```


1. UI Layer (View)
This is what the user sees and interacts with.
Components: InputField, HistoryPanel, ChatPane, MainLayout.
Responsibilities:
Render: Read data from the Store and draw it (render widgets).
Listen: Catch user inputs (clicks, keystrokes).
Stay Dumb: No business logic allowed. If a button is clicked, it doesn't "start a chat", it "announces the intent to start a chat" (creates an Action).
Example: "Send" button clicked -> MainLayout.dispatch(ActionSendMessage(text="Hi")).
2. Actions (Intents)
These are simple message objects (usually dataclass or Pydantic models).
Responsibilities: Carry data about what happened.
Examples:
ActionUserTyped(text="Hello")
ActionAgentStarted(id="123")
ActionErrorOccurred(error="Network fail")
3. Store (State Container)
The central hub.
Responsibilities:
Hold State: The single source of truth for application data (Pydantic model).
Accept Actions: Via the dispatch(action) method.
Protect Data: Uses asyncio.Lock() internally to prevent race conditions.
Notify: When state changes, it triggers subscribers (the UI).
4. Reducer (Pure Logic)
The brain of the system. A pure function.
Input: (Current State, Action)
Output: (New State)
Responsibilities:
Decide how the state changes in response to an action.
Example: ActionUserTyped arrives. The reducer takes the message list, appends the new message, and returns the updated state.
Crucial: No network requests, no print, no sleep. Just data math.
5. Middleware (Side Effects / Async)
The hands of the system. Does the dirty work.
Input: (Action, Store)
Responsibilities:
Listen for Actions.
If an Action requires interacting with the outside world (network, disk, process spawning) — do it here.
When the work is done — create a new Action with the result and dispatch it back to the Store.
Example:
ActionSendMessage arrives.
Middleware sees it, extracts the text, sends it to ConversationRunner.
Waits for response.
Receives "How are you?".
Dispatches ActionAgentResponse(text="I am AI").
6. ConversationRunner (Service)
Part of the Middleware/Services layer.
Responsibilities: Wrapper around the SDK. Knows how to talk to the LLM and manage the agent lifecycle. Knows absolutely nothing about the UI.

malhotra5:
@openhands /codereview-roasted. post back the review as a github comment

malhotra5:
# 🔥 Code Review: State Management Refactor

**Taste Rating:** 🟡 **Acceptable** - Moving in the right direction, but the execution has some fundamental design tensions that need attention.

---

## The Good News First 🐾

This PR tackles a real problem - the god object pattern in `OpenHandsApp` was genuinely painful. The move toward reactive state with `data_bind()` and message bubbling is the right architectural direction for Textual. The removal of the blocking `Future`-based confirmation callback is a significant improvement.

---

## **[CRITICAL ISSUES]** - These Need Rethinking

### 1. **ConversationState is a Lie** 
[openhands_cli/tui/core/state.py, Lines 65-175]

The class is named `ConversationState` but it's actually a `Container` that composes UI widgets. This is a fundamental naming/design mismatch:

```python
class ConversationState(Container):  # 🤔 State that's a Container?
    def compose(self):
        # ... yields UI widgets
```

**The Problem:** You've created a hybrid that's neither pure state nor a proper UI container. The docstring says "Pure reactive state container" but it's composing `ScrollableContent`, `InputAreaContainer`, `WorkingStatusLine`, etc.

**Suggestion:** Either:
- Rename to `ConversationContainer` and accept it's a UI component that owns state, OR
- Extract pure state into a separate non-widget class and have the container reference it

### 2. **ConversationManager is the New God Object**
[openhands_cli/tui/core/conversation_manager.py, Lines 100-592]

You've moved the god object problem from `OpenHandsApp` to `ConversationManager`. It now handles:
- Message routing (8+ `@on` handlers)
- Runner lifecycle management
- Policy management
- Confirmation flow
- Worker spawning
- Visualizer creation

**The Problem:** This is ~500 lines of mixed concerns. The class does too much.

**Suggestion:** Consider splitting:
- `ConversationOperations` - create, switch, pause, condense
- `ConfirmationHandler` - policy and confirmation flow
- Keep `ConversationManager` as a thin coordinator

### 3. **Thread Safety via Exception Catching**
[openhands_cli/tui/core/state.py, Lines 280-292]

```python
def _schedule_update(self, attr: str, value: Any) -> None:
    try:
        self.app.call_from_thread(do_update)
    except RuntimeError:
        # Already on main thread - do update directly
        do_update()
```

**The Problem:** Using exception handling for control flow is a code smell. This pattern assumes `RuntimeError` only means "already on main thread" - fragile assumption.

**Suggestion:** Check thread identity explicitly:
```python
import threading
if threading.current_thread() is threading.main_thread():
    do_update()
else:
    self.app.call_from_thread(do_update)
```

---

## **[IMPROVEMENT OPPORTUNITIES]** - Should Fix

### 4. **Dual State Update Patterns**
[openhands_cli/tui/core/state.py]

You have two ways to update state:
1. Direct assignment: `self.running = False` (line 393)
2. Thread-safe methods: `self.set_running(True)` (line 298)

**The Problem:** Inconsistent API. When do I use which? The `reset_conversation_state()` method uses direct assignment, but `_update_run_status` in runner uses `set_running()`.

**Suggestion:** Pick one pattern. If thread-safety is needed, make ALL updates go through the safe methods and make the reactive vars private.

### 5. **Hidden Dependency for Metrics**
[openhands_cli/tui/core/state.py, Lines 340-355]

```python
def attach_conversation(self, conversation: "BaseConversation") -> None:
    """Attach a conversation for metrics reading."""
    self._conversation = conversation
```

**The Problem:** `ConversationState` holds a reference to `BaseConversation` just to read metrics. This creates a hidden dependency and makes the "state" class know about conversation internals.

**Suggestion:** Have `ConversationManager` push metrics to state instead of state pulling from conversation:
```python
# In ConversationManager
self._state.set_metrics(runner.conversation.state.stats.get_combined_metrics())
```

### 6. **Confirmation Panel Self-Removal Pattern**
[openhands_cli/tui/panels/confirmation_panel.py, Lines 118-140]

The panel posts a message AND removes itself:
```python
self.post_message(ConfirmationDecision(UserConfirmation.ACCEPT))
self._remove_self()
```

**The Problem:** State management is split - the panel removes itself from DOM, but `ConversationManager` manages `pending_action_count`. If the message fails to deliver, you have inconsistent state.

**Suggestion:** Let `ConversationManager` handle removal after processing the decision, or use the reactive pattern fully (panel watches `pending_action_count` and removes itself when it goes to 0).

### 7. **HistorySidePanel's `_previous_is_switching` Hack**
[openhands_cli/tui/panels/history_side_panel.py, Lines 220-235]

```python
self._previous_is_switching: bool = False
# ...
def _on_is_switching_changed(self, is_switching: bool) -> None:
    was_switching = self._previous_is_switching
    self._previous_is_switching = is_switching
```

**The Problem:** You're manually tracking previous state to detect "cancelled switches". This is a workaround for missing state in `ConversationState`.

**Suggestion:** Add explicit state for switch cancellation, or emit a `SwitchCancelled` message.

---

## **[STYLE NOTES]** - Minor Issues

### 8. **Excessive Docstrings**
Multiple files have docstrings that repeat what the code already says clearly. The architecture comments are helpful, but method-level docs like:

```python
def set_running(self, value: bool) -> None:
    """Set the running state. Thread-safe."""
```

...add no value. The method name is self-documenting.

### 9. **Cast to OpenHandsApp**
[openhands_cli/tui/widgets/input_area.py, Lines 95, 117, 150]

```python
app = cast("OpenHandsApp", self.app)
```

**The Problem:** This tight coupling defeats the purpose of decoupling. If `InputAreaContainer` needs `OpenHandsApp` methods, the abstraction is leaking.

---

## **VERDICT:** ✅ **Worth merging with revisions**

The core architectural direction is sound - reactive state, message bubbling, and separation of concerns are the right patterns. However, the execution has created new complexity that should be addressed:

1. **Must fix:** Rename `ConversationState` or restructure it
2. **Must fix:** Fix the thread-safety pattern (don't catch exceptions for control flow)
3. **Should fix:** Consider splitting `ConversationManager`
4. **Should fix:** Unify state update patterns

---

## **KEY INSIGHT:**

> You've successfully killed one god object, but in doing so, you've created a new one (`ConversationManager`) and a hybrid entity (`ConversationState`) that's neither pure state nor pure UI. The next iteration should focus on making these boundaries cleaner.

The reactive pattern with `data_bind()` is excellent - lean into it more. Let state changes drive UI updates, and keep the operation handlers thin.

🐾 *smolpaws approved with requested changes*

malhotra5:
@pashkash we probably won't pursue a state management pattern outside of what textual inherently supports due to time constraints. I'm hoping to get this PR through as soon as possible, but feel free to open an experimental PR for the architecture you've suggested



@enyst I'm happy with the changes so far, I've done a round of iteration with some more notable improvements

1. conversation manager is no longer a god object
2. i've rename conversation state to conversation container to better reflect that it composes child widgets that bind to its data
3. made switching flow and confirmation mode rely on messages rather than directly touching app internals (which follows the pattern introduced in this refactor)


ready for another round of review!

enyst:
@openhands Understand this PR deeply. Tell me how it works and WDYT about it?

openhands_cli/tui/core/conversation_manager.py
```diff
@@ -1,129 +1,292 @@
...
+    def _on_switch_confirmed(self, event: SwitchConfirmed) -> None:
+        """Handle switch confirmation result from modal."""
+        event.stop()
+        self._switch_controller.handle_switch_confirmed(
+            event.target_id,
+            confirmed=event.confirmed,
+        )
+
+    @on(PauseConversation)
+    async def _on_pause_conversation(self, event: PauseConversation) -> None:
```
enyst, line 228: Love these!

enyst (approved):
Thank you so much for the work on this, it looks and reads much better!

I also think that... maybe now, maybe with some docs, agents can have an ~easy time following a similar pattern for new functionality or not breaking this functionality. We shall find out!

xingyaoww:
@OpenHands review diff against main and /remember how the state management works with the app into AGENTS.md

xingyaoww:
@OpenHands please fix the failing actions on PR #421 at branch `refactor/improve-state-management-v2`

Don't change the implementation, just fix the test

xingyaoww:
## Fix for Flaky Snapshot Test (`test_phase2_first_conversation`)

### Root Cause

The Rich library's `export_svg()` method assigns CSS class numbers (r1, r2, r3, etc.) to styles **in the order they are first encountered during rendering**. This order varies between environments (local vs CI) due to differences in how UI elements are composed and rendered.

For example:
- **Local**: `r7=#277dff`, `r8=#706433`, `r9=#222222`
- **CI**: `r7=#706433`, `r8=#222222`, `r9=#277dff`

The colors are identical, but assigned to different class numbers, causing snapshot mismatches.

### Why This Can't Be Fixed at the Source

Looking at Rich's source code (`rich/console.py`), class assignment happens like this:

```python
classes: Dict[str, int] = {}
style_no = 1
# ...
if rules not in classes:
    classes[rules] = style_no
    style_no += 1
```

There's no parameter to control this behavior. The existing `normalize_svg()` in pytest-textual-snapshot only strips the unique ID prefix (`terminal-1234567890-r7` → `terminal-r7`), but doesn't normalize color-to-class mappings.

### Solution

I implemented a `ColorNormalizedSVGExtension` that:
1. Strips the unique ID prefix (like the original `normalize_svg`)
2. Extracts all color definitions and their class names
3. **Remaps class names based on sorted color values** for determinism
4. Updates all references to use the new class names
5. Sorts CSS definitions for consistent ordering

This ensures the same colors always map to the same class numbers after normalization, making snapshots deterministic across environments.

xingyaoww:
@malhotra5 @enyst 

I included this workaround.. LMK if y'all think there's a better solution and i'd be happy to revert this
https://github.com/OpenHands/OpenHands-CLI/pull/421#issuecomment-3867553160

PR #427, opened 2026-01-27
feat(e2e): add mock LLM server and full UI e2e tests

malhotra5:
## Summary

This PR does the following

1. run unit tests and textual snapshot tests in separate CI jobs
2. add e2e tests for snapshot tests 
3. create getters + env overrides for global location constants to simplify test mocking
4. create shareable fixtures for unit tests and dedup/remove inconsistent mocking strategies
5. rename e2e folder to differentiate appropriately from e2e snapshot tests  (binary executable tests vs e2e textual app tests)

For the E2E tests

1. we run the real OpenHands TUI app and simulate user inputs
2. we run a real LLM server which will send back pre-defined LLM responses (mocking responses) 
3. the pre-defined responses are determine by real OpenHands agent trajectories
4. these are simulating real conversations (legitimate tool calls, etc) 

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@e2e-test-mock-llm-server
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,8 +18,10 @@ This repo contains the current CLI UX, including the Textual TUI and a browser-s
 - `.openhands/skills/`: agent guidance for this repo.
 
 ## Setup, Build, and Development Commands
-- install dependencies: `make install`
-- install dev dependencies: `make install-dev`
+This repository uses **uv** for dependency management and running tooling (such as in `Makefile`, CI workflows, and `uv.lock`). Avoid using `pip install ...` directly if possible.
+
+- install dependencies: `make install` (runs `uv sync`)
+- install dev dependencies: `make install-dev` (runs `uv sync --group dev`)
 - install pre-commit hooks: `uv run pre-commit install` (included in `make build`)
 - build (sync + install hooks): `make build`
 - lint (all pre-commit hooks): `make lint`
--- a/openhands_cli/argparsers/util.py
+++ b/openhands_cli/argparsers/util.py
@@ -11,6 +11,7 @@ def add_confirmation_mode_args(
     """
     parser_or_group.add_argument(
         "--always-approve",
+        "--yolo",
         action="store_true",
         help="Auto-approve all actions without asking for confirmation",
     )
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -135,6 +135,20 @@ class TestMainEntryPoint:
         assert kwargs["queued_inputs"] is None
 
     @patch("openhands_cli.tui.textual_app.main")
+    @patch("sys.argv", ["openhands", "--yolo"])
+    def test_main_with_yolo_argument(self, mock_textual_main: MagicMock) -> None:
+        """Test that main() passes always_approve=True with --yolo."""
+        mock_textual_main.side_effect = KeyboardInterrupt()
+
+        main()
+
+        mock_textual_main.assert_called_once()
+        kwargs = mock_textual_main.call_args.kwargs
+        assert kwargs["resume_conversation_id"] is None
+        assert kwargs["always_approve"] is True
+        assert kwargs["queued_inputs"] is None
+
+    @patch("openhands_cli.tui.textual_app.main")
     @patch("sys.argv", ["openhands", "--llm-approve"])
     def test_main_with_llm_approve_argument(self, mock_textual_main: MagicMock) -> None:
         """Test that main() passes llm_approve=True with --llm-approve."""
@@ -158,6 +172,7 @@ class TestMainEntryPoint:
         (["openhands"], None, False, False),
         (["openhands", "--resume", "test-id"], "test-id", False, False),
         (["openhands", "--always-approve"], None, True, False),
+        (["openhands", "--yolo"], None, True, False),
         (["openhands", "--llm-approve"], None, False, True),
         (
             ["openhands", "--resume", "test-id", "--always-approve"],


AGENTS.md
```diff
@@ -55,6 +55,12 @@ Prefer modern typing syntax (`X | None` over `Optional[X]`) in new code.
 - Match test locations to implementation (`tests/` mirrors `openhands_cli/`); add fixtures in `tests/conftest.py` when shared.
 - Run `make test` before PRs.
 
+### E2E Tests with Mock LLM
+- E2E tests in `e2e_tests/` can use `mock_llm_server.py` for deterministic testing without real LLM calls.
+- The mock LLM server provides OpenAI-compatible endpoints with proper tool call format.
+- Use `openai/gpt-4o-mock` as the model name (litellm requires a provider prefix).
+- Note: [llm-mock](https://github.com/piyook/llm-mock) is **not suitable** for OpenHands e2e testing because it doesn't support the `tool_calls` format required by OpenHands agents.
```
malhotra5:
```suggestion
```

enyst:
@openhands Do a /codereview-roasted on this PR.

smolpaws:
## Taste rating: 🟡 Acceptable (but you’re flirting with 🔴 in a couple places)

This PR *mostly* solves real problems (deterministic e2e, less mocking chaos, CI separation). But it also introduces a few “looks fine in CI, hurts later” design choices: weak boundaries, magic behavior, and some coupling you’ll regret.

---

## [CRITICAL ISSUES] (must fix)

### 1) `binary_tests/mock_llm_server.py`: you built a “mock server” that isn’t an OpenAI server, it’s a partial behavioral mimic with undefined failure modes
**Problem:** The handler consumes *any* POST to `/chat/completions`, ignores most request fields, and just pops the next trajectory event. No validation of:
- requested model vs available
- message list shape / tool call formats
- “stream” semantics beyond dumping SSE chunks
- request ordering / idempotency / retries

That means:
- one accidental extra request (client retry, background call, telemetry) silently shifts the replay index → your “deterministic” test becomes a random walk.
- failures won’t be obvious; you’ll just get the “Task completed.” fallback and tests might still pass depending on snapshot content.

**Concrete fix:** make mismatches loud:
- Validate expected request count and shape.
- If you run out of events: return **500** with “trajectory exhausted” (or at least fail-fast unless explicitly configured).
- Optionally match on the last user message content to ensure you’re replaying the right step, not “whatever comes next”.

---

### 2) `HTTPServer.expect_request("").respond_with_handler(handler)` is a footgun
Empty-string expectation is effectively “catch-all” in a way that’s *too implicit*. It’s also not obvious if it matches all paths/methods consistently across versions of `pytest-httpserver`.

**Concrete fix:** register explicit endpoints if possible, or at least comment why the empty path is correct here and add a sanity test that:
- `/v1/chat/completions` hits the handler
- unknown path returns 404

So a dependency upgrade doesn’t silently break your harness.

---

### 3) `openhands_cli/locations.py`: you replaced constants with getters, but didn’t cleanly separate “config” vs “runtime”
This is the right direction for testability, but the current approach is half-measure:
- `get_work_dir()` reads `os.getcwd()` at call time → behavior depends on *current process cwd*, which can change.
- The env override mechanism is scattered (three env vars) and lacks a single source-of-truth object.

**Why it matters:** tests become “set env + hope nobody calls cwd-changing code”. Production behavior can also drift if something `chdir()`s.

**Concrete fix:** introduce one config struct (even a tiny module-level function returning a dataclass) or at minimum:
- make it explicit that `work_dir` is “runtime cwd” and not a stable config path
- consider caching (or passing work dir explicitly into components that need it)

---

## [IMPROVEMENT OPPORTUNITIES] (should fix)

### 4) Mock server concurrency story is confused
You’ve got a lock for `current_index`, good. But:
- streaming responses are built ahead of time and returned as a whole string (fine for tests)
- no per-request correlation: concurrent completions will interleave consumption unpredictably

If e2e ever parallelizes, this breaks instantly.

**Concrete fix:** either:
- document “single-threaded / single client only”, enforce it (reject concurrent requests), or
- add request correlation (match on conversation id / user message) and make replay deterministic.

---

### 5) Default token usage is fake data that may leak into snapshots later
`DEFAULT_USAGE` is constant and unrelated to content. If snapshots ever assert usage (or UI displays it), you’ll bake in nonsense.

**Concrete fix:** either omit `usage` entirely (allowed in many clients) or derive something deterministic from content length.

---

### 6) CI split: good intent, but coverage job now ignores snapshot tests entirely
You moved snapshot tests to a separate job (fine), but coverage is only collected from unit tests (`--ignore=tests/snapshots`). That means coverage % can *inflate* while code exercised only in snapshots regresses.

**Concrete fix:** either:
- run coverage in both jobs and combine, or
- explicitly state “coverage excludes snapshots by design” in CI/job name, and maybe adjust the badge expectations.

---

### 7) Repo structure / naming: “binary_tests” vs “e2e_tests” is still muddy
Renaming helps, but “binary_tests” is not a standard term and doesn’t convey “runs the actual TUI binary”. It reads like “tests for binaries”.

**Concrete fix:** consider `tui_e2e/` or `cli_e2e/` or `integration_tui/`. If you keep `binary_tests`, add one paragraph up top in the README explaining what “binary” means here.

---

## [STYLE NOTES] (minor, but real)

### 8) `TrajectoryResponseConverter._create_default_response()` is called via a “private” method from outside
In `create_request_handler`, you call `converter._create_default_response()` (private method abuse). That’s not “Pythonic freedom”, it’s you admitting the API is wrong.

**Concrete fix:** make it public (`create_default_response`) or handle the “no event left” case in the converter cleanly.

---

## VERDICT
✅ **Probably mergeable** if you tighten the mock server failure behavior.  
❌ **Don’t merge as-is** if you care about “deterministic e2e” meaning anything under retries/concurrency—right now it’s deterministic only in the happy-path fantasy world.

**Key insight:** Your replay mechanism is *index-based*, but real systems are *event-based*. Index-based replay will always be fragile unless you validate/match requests to expected events.

enyst:
> no per-request correlation: concurrent completions will interleave consumption unpredictably

I’m not concerned about concurrency if we don’t use concurrency. The agent is sort of right though, that something feels fragile; and it rings familiar: the old integration tests we had in OH (very old by now) were doing something similar, and to be fair, they proved quite time consuming, because stuff kept breaking them, so humans spent time.

I think… we have gone a really long way amazingly, in what agents can do today. I’m not concerned about this, TBH. Reading through the PR, what I feel is: ok, if something here breaks in a visible way, by all means, let it. The agent will likely make it more resilient next time. Easily.

> consider `tui_e2e/`

+1

> make it public (create_default_response)

Ha! +1 (heavy +1 😅)

enyst (approved):
Seems like a good idea to me! We should try it IMHO.

Issue #444, opened 2026-01-31
Configure timeout using the TUI in the SettingsCard


joseluisll:
Modify the TUI to allow the user to see the configured timeout for the LLM.

The timeout must be a positive integer from 10 to 3600.

The default value must be None.

The setting must be loaded from the agent_settings.json when displaying the SettingsCard.

The setting must be saved to the agent_settings.json when the user clicks the "Save" button of the SettingsCard, only if changed or set to None.

Having this timeout allows for more flexibility for home coders that need to use local LLMs.

This issue will be closed by #445.

Issue #396, opened 2026-01-24
[CLI] Configure max_tokens and max_size using the UI


lowlyocean:
The `agent_settings.json` file saves # of `max_tokens` as null and the `max_size` as 240.

Can we please enhance the TUI to allow configuring these values, and make sure they're saved to the JSON?

joseluisll:
PR #445 solves this issue.

saurya:
Solved in #445 as mentioned above

PR #445, opened 2026-01-31
Add timeout, max_tokens and max_size to the LLM settings configuration card

joseluisll:
Solves the issue #444 . 

Added a timeout field to the SettingsCard. Modified the validator. Reviewed that the field is written to the agent_settings.json file and also read from it. In case of invalid values, the original timeout is preserved. 

Also addresses issue #396: configure max_tokens and max_size via UI.


openhands_cli/tui/modals/settings/components/settings_tab.py
```diff
@@ -96,6 +96,39 @@ def compose(self) -> ComposeResult:
...
+                    with Container(classes="form_group"):
+                        yield Label("LLM Timeout (seconds):", classes="form_label")
+                        yield Input(
+                            placeholder="10–3600 (optional)",
+                            id="timeout_input",
+                            classes="form_input",
+                            # Enabled when API key is entered (same condition as memory)
+                            disabled=True,
+                        )
+
```
malhotra5, line 96: NIT: can we showcase this in advance settings mode instead

tests/snapshots/e2e/__snapshots__/test_creating_agent_plan/TestCreatingAgentPlan.test_creating_agent_plan_conversation[creating_agent_plan].svg
Comment (malhotra5) [inline tests/snapshots/e2e/__snapshots__/test_creating_agent_plan/TestCreatingAgentPlan.test_creating_agent_plan_conversation[creating_agent_plan].svg:1]: Any idea what's going on with the snapshot tests here?

joseluisll:
I will review the PR and apply the changes. I have limited knowledge, but I want to learn an improve at the same time I contribute to this project, which by the way is amazing!. I am using OpenHands-CLI to help me with this work!.

openhands_cli/tui/modals/settings/components/settings_tab.py
joseluisll: Done, moved the Timeout, max_tokens and conderser max_size to the "Advanced".

tests/snapshots/e2e/__snapshots__/test_creating_agent_plan/TestCreatingAgentPlan.test_creating_agent_plan_conversation[creating_agent_plan].svg
Comment (joseluisll) [reply tests/snapshots/e2e/__snapshots__/test_creating_agent_plan/TestCreatingAgentPlan.test_creating_agent_plan_conversation[creating_agent_plan].svg:1]: Yes. The snapshot report says that the TUI interface has changed for 6 test cases. In some cases is just the navigation bar is highlighted when in the original screenshot was not. In other case, the LLM Timeout parameter appears in the advanced  LLM Configuration, when in the original screenshot was not present. In the last case the execution of the hello world script has changed as well: now it does not  report the interpreter, the exit code and the working directory when in the original screenshot that information was present.

[snapshot_report.html](https://github.com/user-attachments/files/25058952/snapshot_report.html)

Comment (joseluisll) [reply tests/snapshots/e2e/__snapshots__/test_creating_agent_plan/TestCreatingAgentPlan.test_creating_agent_plan_conversation[creating_agent_plan].svg:1]: However I have tested manually the hello world run both with the OpenHands-CLI:main unchanged branch and my PR branch, and I now why there is a difference. In my case the command is collapsed, meaning that if I click in the command, I see the same information concerning the interpreter, the exit code and the working directory. I use the setting so that all commands are collapsed by default. Here is the screenshot when I run my branch and then I click on the running command: it expands and show the expected information.

MODIFIED_OPENHANDS-CLI-HELLO-WORD

MY DEFAULT SETTINGS IS COLLAPSED

lowlyocean:
Thanks for tackling this. Can we include num_retries also? Note that settings.json has separate num_retries and timeout for both agent LLM and Condenser LLM

openhands_cli/tui/modals/settings/settings_screen.py
```diff
@@ -199,6 +204,29 @@ def _load_current_settings(self) -> None:
...
         self.memory_select.value = bool(self.current_agent.condenser)
 
+        # Timeout (seconds) – show existing value if set
+        if llm.timeout is not None:
+            self.timeout_input.value = str(llm.timeout)
+        else:
+            self.timeout_input.value = ""
+
+        # Max tokens (optional) – show existing value if set
+        max_output = getattr(llm, "max_output_tokens", None)
```
enyst: I think `max_output_tokens` is a defined attribute, it should always be there, so maybe we don't need getattr?

https://github.com/OpenHands/software-agent-sdk/blob/eb74fa32854320aaf61cfebc0d69d167c8614d7f/openhands-sdk/openhands/sdk/llm/llm.py#L196

enyst: It's just a nit though. Out of curiosity, why output tokens is very interesting, I wonder if it isn't `max_input_tokens` maybe more useful in general?

openhands_cli/tui/modals/settings/utils.py
```diff
@@ -39,6 +46,79 @@ class SettingsFormData(BaseModel):
...
+    @classmethod
+    def validate_max_size(cls, v: str | int | None) -> int | None:
+        """Validate max_size for condenser.
+
+        Must be a positive integer. Returns ``None`` for empty/invalid.
+        """
+        if v is None:
+            return None
+        if isinstance(v, int):
+            return v if v > 0 else None
```
enyst:
```suggestion
            return v if v > 30 else None
```
I think maybe at least 10 would be good to put here, to avoid unlikely, but pointless errors. The condenser is designed to have some space to cut events in half and approximate around the halves. Just a thought, for user friendliness, WDYT?

openhands_cli/tui/modals/settings/utils.py
```diff
@@ -39,6 +46,79 @@ class SettingsFormData(BaseModel):
...
+        if isinstance(v, str):
+            v = v.strip()
+            if v == "":
+                return None
+            if not v.isdigit():
+                return None
+            val = int(v)
+            return val if val > 0 else None
```
enyst: Ditto
```suggestion
            return val if val > 30 else None
```

openhands_cli/tui/modals/settings/settings_screen.py
enyst: @openhands Make an issue to go through the codebase and analyze any getattr and setattr, figure out if we have clear types, and replace them with direct use of the attribute. The goal is to use types, and we have a lot of pydantic classes and dataclasses, we should make use of their expected guarantees.

enyst: I mark this as resolved actually, I set off OpenHands do do a search in the entire codebase and look into these attributes.

enyst:
Thank you, looks good to me!

malhotra5:
@openhands resolve the merge conflicts. note that this PR is from a fork push your changes accordingly

joseluisll:
Thank you for adding this feature.

PR #448, opened 2026-02-03
fix: critic - no module named posthog

brettstewart:
This pull request improves the robustness and optionality of analytics tracking in the critic feedback system by making the PostHog dependency optional and ensuring that analytics events are only sent when PostHog is available and enabled. It also adds a check to only show critic feedback UI elements when the feature is enabled in settings.

**Analytics and PostHog integration:**

* Made the import of the `Posthog` client optional, setting a flag (`POSTHOG_AVAILABLE`) to indicate its availability and preventing errors if the dependency is missing.
* Updated the `send_critic_inference_event` function to return early if PostHog is not available, avoiding unnecessary errors or attempts to send analytics.
* Modified the critic feedback class to only initialize the PostHog client if it is available, and to only send analytics events if PostHog is initialized. [[1]](diffhunk://#diff-dbb0a9e0c5a64f634738a6111d6ae5ef448cc511efddfd4b763a3f682d8a078cR149-R150) [[2]](diffhunk://#diff-dbb0a9e0c5a64f634738a6111d6ae5ef448cc511efddfd4b763a3f682d8a078cL225-R235)

**Feature flag for critic UI:**

* Updated the `richlog_visualizer.py` widget to only add the critic feedback UI if the `enable_critic` setting is enabled, making the feature toggleable via configuration.

**Original Error it addresses**
The CLI Binary has the following error (This does not exist in uv) : ModuleNotFoundError: No module named 'posthog'

**Optional Implementation**
This could be avoided by adding Posthog to the .spec file, but I was unsure whether that was done to avoid bloat in the binary.

### DIFF
--- a/openhands_cli/tui/utils/critic/feedback.py
+++ b/openhands_cli/tui/utils/critic/feedback.py
@@ -5,8 +5,10 @@ from __future__ import annotations
 import uuid
 from typing import TYPE_CHECKING, ClassVar
 
+
 try:
     from posthog import Posthog
+
     POSTHOG_AVAILABLE = True
 except Exception:  # pragma: no cover - optional dependency
     Posthog = None  # type: ignore[assignment]
@@ -17,6 +19,8 @@ from textual.widgets import Button, Static
 
 
 if TYPE_CHECKING:
+    from posthog import Posthog as PosthogClient
+
     from openhands.sdk.critic.result import CriticResult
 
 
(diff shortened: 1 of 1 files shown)


malhotra5:
Thanks @brettstewart 

Maybe we should also modify the pyinstaller spec so that posthog is bundled properly

brettstewart:
> Thanks @brettstewart
> 
> Maybe we should also modify the pyinstaller spec so that posthog is bundled properly

@malhotra5 Agree, I left that out because I wasn't sure if we were trying to avoid bloating the binary. 
With approval, I can add that!

malhotra5:
Please go ahead! If its possible to bundle only portions of the library then that would be most ideal

brettstewart:
> Please go ahead! If its possible to bundle only portions of the library then that would be most ideal

@malhotra5 added PostHog to hiddenimports and metadata (and removed it from excludes) to force PyInstaller to bundle the optional dependency without pulling in extra submodules

Issue #513, opened 2026-02-17
crash on first start


montvid:
crash on first start after uv tool install openhands --python 3.12
OpenHands CLI 1.12.1
Python 3.13.5
uv 0.10.3
`openhands
╭──────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────╮
│ /home/kvb/.local/share/uv/tools/openhands/lib/python3.12/site-packages/openhands_cli/tui/modals/settings/settings_ │
│ screen.py:142 in on_show                                                                                           │
│                                                                                                                    │
│   139 │   │   # Only reload if we don't have current settings loaded                                               │
│   140 │   │   # This prevents unnecessary clearing when returning from modals                                      │
│   141 │   │   if not self.current_agent:                                                                           │
│ ❱ 142 │   │   │   self._clear_form()                                                                               │
│   143 │   │   │   self._load_current_settings()                                                                    │
│   144 │   │   │   self._update_advanced_visibility()                                                               │
│   145 │   │   │   self._update_field_dependencies()                                                                │
│                                                                                                                    │
│ ╭──────── locals ─────────╮                                                                                        │
│ │ self = SettingsScreen() │                                                                                        │
│ ╰─────────────────────────╯                                                                                        │
│                                                                                                                    │
│ /home/kvb/.local/share/uv/tools/openhands/lib/python3.12/site-packages/openhands_cli/tui/modals/settings/settings_ │
│ screen.py:155 in _clear_form                                                                                       │
│                                                                                                                    │
│   152 │   │   self.custom_model_input.value = ""                                                                   │
│   153 │   │   self.base_url_input.value = ""                                                                       │
│   154 │   │   self.mode_select.value = "basic"                                                                     │
│ ❱ 155 │   │   self.provider_select.value = Select.BLANK                                                            │
│   156 │   │   self.model_select.value = Select.BLANK                                                               │
│   157 │   │   self.memory_select.value = False                                                                     │
│   158                                                                                                              │
│                                                                                                                    │
│ ╭──────── locals ─────────╮                                                                                        │
│ │ self = SettingsScreen() │                                                                                        │
│ ╰─────────────────────────╯                                                                                        │
│                                                                                                                    │
│ /home/kvb/.local/share/uv/tools/openhands/lib/python3.12/site-packages/textual/widgets/_select.py:591 in           │
│ _validate_value                                                                                                    │
│                                                                                                                    │
[...]

PR #514, opened 2026-02-17
fix: Clear values of settings modal using .clear() instead of setting as blank

csmith49:
Textual 8.0.0 [changed](https://github.com/Textualize/textual/blob/main/CHANGELOG.md) how `Select.BLANK` is interpreted. Instead of being a `NoSelection` instance, it becomes `False`, which is causing crashes on first startup for some users (see #513).

This PR changes explicit assignments to `Select.BLANK` to use `.clear()` instead, which sets the value to `NoSelection`. It also updates comparisons to `Select.BLANK` to use `isinstance(..., NoSelection)`.

Another fix might have been to replace `Select.BLANK` to `Select.NULL`, but the changes done here should work for Textual versions pre- and post-8.0.0 without issue.

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@fix/settings-clearing
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
   "openhands-workspace==1.11.1",
   "prompt-toolkit>=3",
   "rich<14.3.0",
-  "textual==8.0.0",
+  "textual>=8.0.0,<9.0.0",
   "typer>=0.17.4",
   "agent-client-protocol>=0.7.0,<0.8.0",
   "pydantic>=2.7",


malhotra5 (approved):
Thank you so much! 

Do you also mind pinning the textual version in poetry as to prevent this from happening in the future

malhotra5:
@openhands please pin the version of textual in pyproject and regenerate the lock file

PR #528, opened 2026-02-21
fix(Makefile): help and pre-commit

VascoSch92:
A couple of things:
- `make lint` was wrong (it was running the `pre-commit`)
- I added `make pre-commit` to run the `pre-commit`
- Restyling of `make help` (it was incomplete)

Screenshot 2026-02-21 at 23 26 47


VascoSch92:
> LGTM, can we make sure that AGENTS.md is consistent with it?

It is just one line change for he make pre-commit. (Which actually is not needed)

Otherwise, it is consistent.

PR #534, opened 2026-02-23
fix: remove unused runtime image pull and env var from gui_launcher

neubig:
## Summary

This PR removes the unnecessary runtime image pull and unused `SANDBOX_RUNTIME_CONTAINER_IMAGE` env var from `gui_launcher.py`. The CLI was pulling `docker.openhands.dev/openhands/runtime:latest-nikolaik` and setting an environment variable that the V1 OpenHands app completely ignores.

## What Changed

- Removed the `docker pull` command for the unused runtime image
- Removed the `SANDBOX_RUNTIME_CONTAINER_IMAGE` env var (V0-era, ignored by V1 app)
- Let the app use its built-in default agent-server image

## Why This Is Safe

The V1 OpenHands app:
1. Uses `AGENT_SERVER_IMAGE_REPOSITORY` and `AGENT_SERVER_IMAGE_TAG` env vars (not `SANDBOX_RUNTIME_CONTAINER_IMAGE`)
2. Has a hardcoded default: `ghcr.io/openhands/agent-server:010e847-python`
3. The default is updated during SDK version bumps to ensure compatibility

## Root Cause of Linux Timeout Issue (OpenHands/OpenHands#12528)

**This PR does NOT fix the actual timeout bug** - it just cleans up unused code.

The real bug is in `openhands/app_server/sandbox/sandbox_service.py` at lines 116-121:

```python
if sandbox.status == SandboxStatus.RUNNING:
    if httpx_client and sandbox.exposed_urls:
        if await self._check_agent_server_alive(sandbox, httpx_client):
            return sandbox
    else:
        return sandbox  # BUG: Returns immediately even if exposed_urls is empty!
```

On Linux (native Docker), there's a race condition:
1. Container starts, Docker reports status as `running`
2. But port bindings (`NetworkSettings.Ports`) aren't populated yet
3. `exposed_urls` is empty `[]` (falsy in Python)
4. The `else` branch executes and returns immediately
5. Later, when trying to use the sandbox, there are no URLs to connect to

On macOS (Docker Desktop), port bindings are ready faster, masking this bug.

**The fix should be in OpenHands main repo**: Don't return early if `exposed_urls` is empty - continue polling until URLs are populated.

## Testing

Tested both with and without this change - the app starts successfully either way because `SANDBOX_RUNTIME_CONTAINER_IMAGE` is ignored by V1.

## Related

- Related to OpenHands/OpenHands#12528 (cleanup, not fix)
- The actual fix needs to be in the OpenHands main repo

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@fix/use-v1-agent-server-env-vars
```


openhands_cli/gui_launcher.py
```diff
@@ -106,19 +106,22 @@ def launch_gui_server(mount_cwd: bool = False, gpu: bool = False) -> None:
 
     # Get the current version for the Docker image
     version = get_openhands_version()
-    runtime_image = f"docker.openhands.dev/openhands/runtime:{version}-nikolaik"
+    # Agent server image repository and tag (used by sandbox service)
+    agent_server_image_repository = "docker.openhands.dev/openhands/runtime"
```
enyst: Is this accurate? I thought we use what shows in the PR descriptions, for example:

Image

neubig: Yes, this is accurate. The V1 OpenHands app uses `AGENT_SERVER_IMAGE_REPOSITORY` and `AGENT_SERVER_IMAGE_TAG` (defaulting to `ghcr.io/openhands/agent-server:...`), not `SANDBOX_RUNTIME_CONTAINER_IMAGE`.

I verified in the codebase - `SANDBOX_RUNTIME_CONTAINER_IMAGE` is not referenced anywhere in OpenHands main (V1). It was a V0-era env var.

The screenshot shows the runtime image that was being pulled (nikolaik), but this image is not used by V1 - the CLI was pulling it unnecessarily. The actual agent-server image used by the app is specified in `openhands/app_server/sandbox/sandbox_spec_service.py`.

PR #536, opened 2026-02-24
fix(acp): Only advertise OAuth when not already authenticated

### LINKED ISSUE #535: ACP server always advertises OAuth authentication even when local settings exist
## Description

When using the OpenHands CLI ACP server with JetBrains IDEs (IntelliJ, etc.), users see an "Authenticate with OpenHands Cloud" button in the AI Chat window even when they have already configured local LLM settings in `~/.openhands/agent_settings.json`.

## Root Cause

In `openhands_cli/acp_impl/agent/base_agent.py`, the `initialize()` method unconditionally returns OAuth as an available authentication method (line 231 comment says "Always configure auth method"):

```python
auth_methods = [
    AuthMethod(
        description="Authenticate through agent",
        id="oauth",
        name="OAuth with OpenHands Cloud",
        field_meta={"type": "agent"},
    ),
]
```

This causes JetBrains AI Assistant to show the authentication button even when it is not needed.

For the **local agent** (`LocalOpenHandsACPAgent`), the `_is_authenticated()` method returns `True` when `agent_settings.json` exists. However, `_is_authenticated()` is only called in `new_session()`, not in `initialize()`.

## Expected Behavior

When running in local mode (no `--cloud` flag) with `~/.openhands/agent_settings.json` already configured, the ACP server should not advertise OAuth authentication. The `auth_methods` field in `InitializeResponse` should be empty.

## Proposed Fix

Modify `initialize()` to check if authentication is actually needed:

```python
# Only advertise auth methods if not already authenticated
auth_methods = []
if not await self._is_authenticated():
    auth_methods = [
        AuthMethod(
            description="Authenticate through agent",
            id="oauth",
            name="OAuth with OpenHands Cloud",
            field_meta={"type": "agent"},
        ),
    ]
```

## Steps to Reproduce

1. Install OpenHands CLI and configure LLM settings via `openhands` → `/settings`
2. Configure JetBrains IDE with `~/.jetbrains/acp.json` pointing to `openhands acp`
3. Open JetBrains IDE with AI Assistant enabled
4. Select OpenHands agent in AI Chat
5. Observe "Authenticate with OpenHands Cloud" button appears even though local settings exist

## Environment

- OpenHands CLI version: latest
- JetBrains IDE: IntelliJ IDEA 25.3+
- OS: macOS

jpshackelford:
## Summary

Fixes #535

When using the OpenHands CLI ACP server with JetBrains IDEs (IntelliJ, etc.), users see an "Authenticate with OpenHands Cloud" button in the AI Chat window even when they have already configured local LLM settings in `~/.openhands/agent_settings.json`.

## Root Cause

In `openhands_cli/acp_impl/agent/base_agent.py`, the `initialize()` method unconditionally returned OAuth as an available authentication method, regardless of whether authentication was actually needed.

## Changes

- Modified `initialize()` in `base_agent.py` to check `_is_authenticated()` before including auth methods in the response
- When already authenticated (local agent has `agent_settings.json`, or cloud agent has valid API key), `auth_methods` is returned as an empty list
- Updated and added tests to verify the conditional behavior

## Expected Behavior After Fix

- **Local mode with existing settings**: No auth button shown (auth_methods is empty)
- **Local mode without settings**: Auth button shown (auth_methods contains OAuth)
- **Cloud mode with valid API key**: No auth button shown
- **Cloud mode without API key**: Auth button shown

## Commands Run

```bash
make lint        # All checks passed
make test        # All 1257 tests passed
```

## Testing

- Added `test_initialize_no_auth_methods_when_authenticated` - verifies empty auth_methods when authenticated
- Added `test_initialize_auth_methods_when_not_authenticated` - verifies OAuth is returned when not authenticated
- Existing tests continue to pass

@jpshackelford can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@openhands/fix-acp-auth-advertise-535
```

### DIFF
--- a/openhands_cli/acp_impl/agent/base_agent.py
+++ b/openhands_cli/acp_impl/agent/base_agent.py
@@ -228,15 +228,17 @@ class BaseOpenHandsACPAgent(ACPAgent, ABC):
         """Initialize the ACP protocol."""
         logger.info(f"Initializing ACP with protocol version: {protocol_version}")
 
-        # Always configure auth method
-        auth_methods = [
-            AuthMethod(
-                description="Authenticate through agent",
-                id="oauth",
-                name="OAuth with OpenHands Cloud",
-                field_meta={"type": "agent"},
-            ),
-        ]
+        # Only advertise auth methods if not already authenticated
+        auth_methods: list[AuthMethod] = []
+        if not await self._is_authenticated():
+            auth_methods = [
+                AuthMethod(
+                    description="Authenticate through agent",
+                    id="oauth",
+                    name="OAuth with OpenHands Cloud",
+                    field_meta={"type": "agent"},
+                ),
+            ]
 
         return InitializeResponse(
             protocol_version=protocol_version,
--- a/openhands_cli/acp_impl/agent/local_agent.py
+++ b/openhands_cli/acp_impl/agent/local_agent.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import logging
+import os
 from pathlib import Path
 from typing import Any
 from uuid import UUID
@@ -65,15 +66,48 @@ class LocalOpenHandsACPAgent(BaseOpenHandsACPAgent):
         return "local"
 
     async def _is_authenticated(self) -> bool:
-        """Check if agent settings already exist for is_authenticated status.
+        """Check if agent settings exist or AWS credentials are configured.
 
-        For local agent, authentication is considered complete if agent specs exist.
+        For local agent, authentication is considered complete if:
+        1. Agent specs exist (agent_settings.json), OR
+        2. AWS credentials are available (for Bedrock users without agent_settings.json)
         """
         try:
             load_agent_specs()
             return True
         except MissingAgentSpec:
-            return False
+            # No agent_settings.json - check if user has AWS credentials for Bedrock
+            return self._has_aws_credentials()
+
+    def _has_aws_credentials(self) -> bool:
+        """Check if AWS credentials are available for Bedrock authentication.
+
+        This detects users who intend to use AWS Bedrock but haven't yet created
+        an agent_settings.json file. We check for:
+        1. AWS environment variables (explicit credentials or profile)
+        2. AWS credentials/config files (~/.aws/)
+
+        Note: This does NOT validate that credentials are correct - it only
+        detects intent to use AWS-based authentication.
+        """
+        # Check AWS environment variables
+        aws_env_vars = [
+            "AWS_ACCESS_KEY_ID",
+            "AWS_SECRET_ACCESS_KEY",
+            "AWS_PROFILE",
+            "AWS_DEFAULT_PROFILE",
+        ]
+        if any(os.environ.get(var) for var in aws_env_vars):
+            logger.debug("AWS credentials detected via environment variables")
+            return True
+
+        # Check for AWS credentials/config files
+        aws_dir = Path.home() / ".aws"
+        if (aws_dir / "credentials").exists() or (aws_dir / "config").exists():
+            logger.debug("AWS credentials detected via ~/.aws/ files")
+            return True
+
+        return False
 
     def _cleanup_session(self, session_id: str) -> None:
         """Clean up resources for a session (no-op for local agent)."""
--- a/tests/acp/test_agent.py
+++ b/tests/acp/test_agent.py
@@ -28,20 +28,37 @@ def acp_agent(mock_connection):
 
 
 @pytest.mark.asyncio
-async def test_initialize(acp_agent):
-    """Test agent initialization always returns auth method."""
-    response = await acp_agent.initialize(
-        protocol_version=1,
-        client_info=Implementation(name="test-client", version="1.0.0"),
-    )
+async def test_initialize_returns_auth_methods_when_not_authenticated(acp_agent):
+    """Test agent initialization returns auth method when not authenticated."""
+    with patch.object(acp_agent, "_is_authenticated", return_value=False):
+        response = await acp_agent.initialize(
+            protocol_version=1,
+            client_info=Implementation(name="test-client", version="1.0.0"),
+        )
 
-    assert response.protocol_version == 1
-    assert isinstance(response.agent_capabilities, AgentCapabilities)
-    assert response.agent_capabilities.load_session is True
-    # Auth method is always returned for OAuth authentication
-    assert len(response.auth_methods) == 1
-    assert response.auth_methods[0].id == "oauth"
-    assert response.auth_methods[0].field_meta == {"type": "agent"}
+        assert response.protocol_version == 1
+        assert isinstance(response.agent_capabilities, AgentCapabilities)
+        assert response.agent_capabilities.load_session is True
+        # Auth method returned when not authenticated
+        assert response.auth_methods is not None
+        assert len(response.auth_methods) == 1
+        assert response.auth_methods[0].id == "oauth"
+        assert response.auth_methods[0].field_meta == {"type": "agent"}
+
+
+@pytest.mark.asyncio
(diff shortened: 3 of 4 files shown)


jpshackelford:
I have put this back into draft because the PR doesn't properly consider the case of bedrock.  I believe that one authentication mode for Bedrock is the boto3 is used to take authentication detail from an aws profile. Actually I think with Bedrock and LiteLLM there are three distinct modes of authentication:

1. AWS Profile
2. Combination of `AWS_ACCESS_KEY_ID`, `AWS_ACCESS_KEY_ID`, and `AWS_REGION_NAME`
3. New short lived / long lived tokens.

@openhands Please do look into the version of LiteLLM we are using in the CLI or transitively via `OpenHands/software-agent-sdk` and confirm or deny this claim.

openhands_cli/acp_impl/agent/local_agent.py
```diff
@@ -65,15 +66,48 @@ def agent_type(self) -> AgentType:
...
+        1. Agent specs exist (agent_settings.json), OR
+        2. AWS credentials are available (for Bedrock users without agent_settings.json)
         """
         try:
             load_agent_specs()
             return True
         except MissingAgentSpec:
-            return False
+            # No agent_settings.json - check if user has AWS credentials for Bedrock
+            return self._has_aws_credentials()
```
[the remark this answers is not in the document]
jpshackelford: Yes. This is verified in a customer environment.

openhands_cli/acp_impl/agent/local_agent.py
```diff
@@ -65,15 +66,48 @@ def agent_type(self) -> AgentType:
...
+        ]
+        if any(os.environ.get(var) for var in aws_env_vars):
+            logger.debug("AWS credentials detected via environment variables")
+            return True
+
+        # Check for AWS credentials/config files
+        aws_dir = Path.home() / ".aws"
+        if (aws_dir / "credentials").exists() or (aws_dir / "config").exists():
+            logger.debug("AWS credentials detected via ~/.aws/ files")
+            return True
```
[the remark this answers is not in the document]
jpshackelford: We do not want to create a dependency on boto3 library when only Bedrock users will need it. Invalid creds is fine. Customer can fix them. We do not want to try to authenticate against OpenHands OAUTH of user I providing their own AWS creds.

tests/acp/test_agent.py
```diff
@@ -748,3 +765,116 @@ async def test_new_session_no_replay_for_new_conversation(mock_connection, tmp_p
...
+    """Tests for _has_aws_credentials method."""
+
+    def test_has_aws_credentials_with_access_key_env(self, acp_agent, monkeypatch):
+        """Test detection of AWS_ACCESS_KEY_ID environment variable."""
+        monkeypatch.setenv("AWS_ACCESS_KEY_ID", "test-key")
+        assert acp_agent._has_aws_credentials() is True
+
+    def test_has_aws_credentials_with_secret_key_env(self, acp_agent, monkeypatch):
+        """Test detection of AWS_SECRET_ACCESS_KEY environment variable."""
+        monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", "test-secret")
```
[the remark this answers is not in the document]
jpshackelford: Not theoretical. Real production issue with real customer.

openhands_cli/acp_impl/agent/base_agent.py
```diff
@@ -228,15 +228,17 @@ async def initialize(
...
+        auth_methods: list[AuthMethod] = []
+        if not await self._is_authenticated():
+            auth_methods = [
+                AuthMethod(
+                    description="Authenticate through agent",
+                    id="oauth",
+                    name="OAuth with OpenHands Cloud",
+                    field_meta={"type": "agent"},
+                ),
+            ]
```
malhotra5, line 241: Just to note, I think the auth methods should always be declared (not conditionally the way it is now)

To request the user to authenticate, the local (or remote) agent raises `RequestError.auth_required` (here we should verify whether local settings are already configured, and raise conditionally)

malhotra5: It seems we already do that, so unclear why we're seeing this bug

jpshackelford: The issue is that JetBrains interprets non-empty `auth_methods` in the `InitializeResponse` as "authentication is required" and blocks the chat session from starting until the user clicks the auth button.

So even though we correctly raise `RequestError.auth_required` in `new_session()` when needed, the JetBrains client never gets that far - it sees `auth_methods` containing OAuth and stops the user at the gate with the auth button.

By conditionally returning empty `auth_methods` when the user is already authenticated, the JetBrains client allows the session to proceed normally without requiring the user to click through unnecessary auth UI.

openhands_cli/acp_impl/agent/local_agent.py
```diff
@@ -65,15 +66,48 @@ def agent_type(self) -> AgentType:
...
+
+        Note: This does NOT validate that credentials are correct - it only
+        detects intent to use AWS-based authentication.
+        """
+        # Check AWS environment variables
+        aws_env_vars = [
+            "AWS_ACCESS_KEY_ID",
+            "AWS_SECRET_ACCESS_KEY",
+            "AWS_PROFILE",
+            "AWS_DEFAULT_PROFILE",
```
enyst, line 98: This tiny detail feels like maybe it should be taken from the SDK. We just fixed an issue with it:
- https://github.com/OpenHands/software-agent-sdk/pull/2195

And there's another PR for Vertex coming up.

The user may have had these set, but also could change LLM.

How about we expose publicly a function in the SDK?

Though I don't fully understand why there is no agent_settings.json at all, should we make one (without LLM or with null LLM)?

jpshackelford: > I don't fully understand why there is no agent_settings.json at all

My understanding is that when OpenHands CLI first starts up it can detect and create agent_settings.json from environment variable. I am not trying to change that behavior, just allow acp to startup properly w/ JetBrains whether previously configured or not.

jpshackelford: > This tiny detail feels like maybe it should be taken from the SDK

 1.  The SDK's infer_litellm_provider() detects if a model is Bedrock-based, but it's not publicly exported and requires a model name
 2.  Our _has_aws_credentials() check is about detecting AWS credentials in the environment - this is orthogonal to what the SDK does
 3.  The SDK doesn't have a public "are AWS credentials available" function - that would need to be added
 
 Since this is a blocking issue for user now I'd like to fix in CLI and move to `software-agent-sdk` as a refactor when delay of release process for that isn't impactful.

enyst: OK, it's a very localized thing, should be easy to change, so if Rohit is ok with it all good

openhands_cli/acp_impl/agent/base_agent.py
malhotra5: Sounds good thanks for the breakdown! It seems not all IDEs are interpreting the protocol the exact same way (AFIAK Zed operates on the former behavior)

Marking this as resolved!

malhotra5 (approved):
LGTM thank you!

jpshackelford:
I am holding off merging until I get successful end-to-end test with IntelliJ IDEA. There appears to be another problem and it looks like it may be on the IntelliJ side.

jpshackelford:
I am going to test at customer site with just https://github.com/OpenHands/OpenHands-CLI/pull/543 as this PR (#536)  may be chasing a red herring. We'll see.

PR #538, opened 2026-02-24
Allow Vertex models in CLI without requiring LLM API key

aking13:
## Summary
- allow `vertex_ai` / `vertex_ai_beta` models to be configured in CLI settings without forcing an API key
- allow `--override-with-envs` / headless agent creation for Vertex models without `LLM_API_KEY`
- keep API key requirement for non-Vertex providers/models

## Why
Vertex authentication is handled via GCP credentials (for example `GOOGLE_APPLICATION_CREDENTIALS`, `VERTEXAI_PROJECT`, `VERTEXAI_LOCATION`), so requiring a fake API key in CLI was incorrect and confusing.

## Changes
- `openhands_cli/tui/modals/settings/utils.py`
  - add Vertex provider/model detection
  - make API key validation conditional (skip for Vertex)
- `openhands_cli/stores/agent_store.py`
  - add `model_requires_api_key()` helper
  - relax headless env validation for Vertex models
  - permit agent creation with `api_key=None` for Vertex models
- tests
  - add settings tests for Vertex-without-key behavior
  - add env override/headless tests for Vertex-without-`LLM_API_KEY`

## Validation
- `uv run ruff check openhands_cli/tui/modals/settings/utils.py openhands_cli/stores/agent_store.py tests/tui/modals/settings/test_settings_utils.py tests/stores/test_env_llm_overrides.py`
- `uv run pytest -q tests/tui/modals/settings/test_settings_utils.py tests/stores/test_env_llm_overrides.py`
  - result: `59 passed`

## Notes
This PR addresses CLI-side validation/agent creation behavior. Runtime import errors for Vertex still require the environment to include `google-cloud-aiplatform` (LiteLLM dependency).

### DIFF
--- a/openhands_cli/stores/agent_store.py
+++ b/openhands_cli/stores/agent_store.py
@@ -127,6 +127,22 @@ ENV_LLM_BASE_URL = "LLM_BASE_URL"
 ENV_LLM_MODEL = "LLM_MODEL"
 
 
+def model_requires_api_key(model: str | None) -> bool:
+    """Return whether an LLM model requires an API key env var.
+
+    Vertex AI models are authenticated via GCP credentials (ADC/service account),
+    so LLM_API_KEY should not be mandatory for those model prefixes.
+    """
+    if not model:
+        # Be conservative when unknown.
+        return True
+
+    lowered = model.lower()
+    return not (
+        lowered.startswith("vertex_ai/") or lowered.startswith("vertex_ai_beta/")
+    )
+
+
 class MissingEnvironmentVariablesError(Exception):
     """Raised when required environment variables are missing for headless mode.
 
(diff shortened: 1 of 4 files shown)


enyst (commented):
Thank you for looking into this, @aking13 ! Just for information, interestingly we just merged a PR fixing the same thing for Bedrock:
https://github.com/OpenHands/software-agent-sdk/pull/2195

WDYT, could that work for Vertex without API key?

aking13:
> Thank you for looking into this, @aking13 ! Just for information, interestingly we just merged a PR fixing the same thing for Bedrock: [OpenHands/software-agent-sdk#2195](https://github.com/OpenHands/software-agent-sdk/pull/2195)
> 
> WDYT, could that work for Vertex without API key?

I'm not sure I follow. That fix correctly doesn't forward an api_key, but it does not prevent the CLI from hard requiring an API in settings and for headless flows for Vertex, when in reality no API Key is needed at all for everything to work with Vertex. Essentially today Vertex users have to type in some random non-null gibberish key to get Vertex LLMs to work in CLI and this PR removes that experience.

PR #545, opened 2026-02-25
tui: change 'command palette' to 'settings' in plan panel description

### LINKED ISSUE #525: Should ^p palette be ^p settings ?
I remember that settings was open with /settings, and ^p only opened the theme selector.
Took me a while to realize that settings are now under ^p

juanmichelini:
## Description

This PR changes the text "command palette" to "settings" in the plan panel description text in the CLI settings tab. This is a minimal change that makes the terminology more consistent with how users refer to the Ctrl+P feature.

Fixes #525

## What changed

Changed line 79 in `openhands_cli/tui/modals/settings/components/cli_settings_tab.py`:
- Before: "You can toggle it anytime via the command palette."
- After: "You can toggle it anytime via settings."

## Testing

- ✅ `make lint` - All checks passed
- ✅ `make test` - All 1260 tests passed

## Notes

This is a minimal text-only change with no functional modifications. The change makes the UI text more consistent with how users refer to accessing settings via Ctrl+P.

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@openhands/change-palette-to-settings
```

### DIFF
--- a/openhands_cli/tui/textual_app.py
+++ b/openhands_cli/tui/textual_app.py
@@ -43,6 +43,7 @@ from typing import ClassVar
 
 from textual import events, getters, on
 from textual.app import App, ComposeResult, SystemCommand
+from textual.binding import Binding
 from textual.containers import Horizontal
 from textual.screen import Screen
 from textual.widgets import Footer, Input, TextArea
@@ -94,6 +95,7 @@ class OpenHandsApp(CollapsibleNavigationMixin, App):
 
     # Key bindings
     BINDINGS: ClassVar = [
+        Binding("ctrl+p", "command_palette", "Settings", show=False, priority=True),
         ("ctrl+l", "toggle_input_mode", "Toggle single/multi-line input"),
         ("ctrl+o", "toggle_cells", "Toggle Cells"),
         ("ctrl+j", "submit_textarea", "Submit multi-line input"),


juanmichelini:
@openhands I tried this but it converted "^p" into "settings", not "palette" into "settings"
See 
image

juanmichelini:
@openhands now the lower right corner is correct, but "^p settings" was also unnecessary added in between "^q quit the application" and "^o toggle cells" see image:
image

juanmichelini:
I've fixed the issue by adding a proper `Binding` object for the command palette.

**The Problem:**
- The footer has two separate fields: **key display** (e.g., "^p") and **description** (e.g., "palette")
- `COMMAND_PALETTE_DISPLAY` only controls the key display, not the description
- Simply adding `("ctrl+p", "command_palette", "Settings")` to BINDINGS created a duplicate footer entry

**The Solution:**
Add a full `Binding` object with the exact same parameters Textual uses (`show=False`, `priority=True`), but with description "Settings":

```python
Binding("ctrl+p", "command_palette", "Settings", show=False, priority=True)
```

This prevents Textual from adding its default "palette" binding (since it detects ours exists) while avoiding duplicate footer entries.

**Result:**
- ✅ Footer shows "^p Settings" (not "^p palette")
- ✅ No duplicate binding entries  
- ✅ Plan panel description says "settings" (from original commit)
- ✅ `make lint` passes

Commit: 043a759

juanmichelini:
@malhotra5 none of the fixes seem satisfying to me, thoughts?

malhotra5:
@juanmichelini maybe we can get rid of the command palette all together (no binding whatsoever)

If we want to access anything, it should be via a slash command

juanmichelini:
Cool, closing this.

PR #546, opened 2026-02-25
tui: implement conditional autoscrolling

### LINKED ISSUE #520: Autoscrolling should be disabled when manually scrolling
Auto scrolling should be enabled when the scrolling position shows the newest message.
If the user scrolls so that the newest message is not shown, then autoscroll should be disabled.

This is useful so that you always get new messsages when you care about them, but you are also able to scroll towards and old message and copy it without the scrolling suddenly jumping.

juanmichelini:
## Summary

This PR implements conditional autoscrolling in the TUI conversation view to address issue #520. The feature ensures that autoscrolling only occurs when the user is already viewing the bottom of the conversation, allowing users to scroll up and read or copy old messages without being interrupted by new content jumping the view back to the bottom.

### Implementation
- Modified `_add_widget_to_ui()` in `richlog_visualizer.py` to check `is_vertical_scroll_end` before calling `scroll_end()`
- Modified `_update_widget_in_ui()` in `richlog_visualizer.py` with the same conditional scroll logic
- Added detailed comments explaining the autoscroll behavior

## Testing
- All existing tests pass (67/67 tests in `test_richlog_visualizer.py`)
- Linting passes with `make lint`
- Tested manually

### Before
- New messages always triggered autoscroll to bottom
- Users couldn't scroll up to read/copy old messages without being interrupted

### After
- Autoscroll only happens when user is already at the bottom (viewing newest messages)
- Users can scroll up to read/copy old messages without interruption
- Autoscroll automatically re-enables when user scrolls back to bottom

Closes #520

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@openhands/autoscroll-on-scroll-end
```


juanmichelini:
This is tested manually as well.

juanmichelini:
@openhands remove the mock tests and fix the 🟢 Nit: These two comment lines are duplicated in both _add_widget_to_ui() and _update_widget_in_ui(). The code is already self-documenting with is_vertical_scroll_end, so this could be condensed to a single line or removed entirely per repo guidelines ("minimize comments that repeat what code shows").

we already test manually, and don't want to retest so this change should only remove those mock tests and those particular comments. everything else should remain ghte same.

openhands_cli/tui/widgets/richlog_visualizer.py
```diff
@@ -290,8 +290,8 @@ def on_event(self, event: Event) -> None:
     def _add_widget_to_ui(self, widget: "Widget") -> None:
         """Add a widget to the UI (must be called from main thread)."""
         self._container.mount(widget)
-        # Automatically scroll to the bottom to show the newly added widget
-        self._container.scroll_end(animate=False)
+        if self._container.is_vertical_scroll_end:
```
malhotra5, line 293: Hmm I would think we need to check the scroll position before mounting the new widget. After mounting we may not be at the end?

malhotra5: Hmm seems to work still, lets merge this

PR #403, opened 2026-01-26
feat(tui): conversation history rehydration with pagination (#204)

pashkash:
## Summary
- Implement conversation history rehydration when resuming or switching conversations
- Add `ConversationPane` smart container for per-conversation UI caching
- Add `ConversationHistoryManager` for pagination state management
- Add `ConversationSessionManager` for runner/pane caching (instant switching)
- Add "Load More" button for loading older history with scroll position preservation

## Changes
- New files: `conversation_pane.py`, `conversation_history.py`, `conversation_history_manager.py`, `conversation_session_manager.py`, `system_splash_pane.py`
- Updated `conversation_switcher.py` to use cached panes
- Updated `textual_app.py` for dynamic pane management
- Updated `richlog_visualizer.py` for delegation message rendering

## Test plan
- [x] `make lint`
- [x] `make test` (490 TUI tests passed)
- [x] `uv run pytest tests/snapshots/` (13 snapshots passed)
- [x] Manual testing: conversation switching, history loading, Load More button

Closes #204

PR #562, opened 2026-03-02
tui: replay historical events on resume/switch

ironbelly:
tui: replay historical events on resume/switch (#v0.01)
When resuming a conversation via --resume <id> or switching via /history,
historical events are now replayed through the ConversationVisualizer so
the user sees the full conversation history in the TUI.

Changes:
- Add ConversationVisualizer.replay_events() for side-effect-free replay
  (skips critic handling, telemetry, plan panel refreshes; renders user
  messages inline)
- Add ConversationRunner.replay_historical_events() with idempotence guard
- Trigger replay in RunnerRegistry.get_or_create() for newly created runners
- Add 7 unit tests covering replay order, empty/idempotent cases, and
  registry integration

**PR #403 Comparison and Approach debate:** : https://github.com/IronbellyOrg/Ironhands-CLI_dev/blob/main/Releases/Complete/v0.01-MessageHistory/adversarial-analysis.md

Verified: make lint, make test (1268 passed), make test-snapshots (59 passed)

fix: resolve 5 message-history bugs for v0.02 patch release (#v0.02)
  BUG-001 (High): --resume shows blank screen instead of conversation history
  - Add ensure_runner() to ConversationManager that eagerly calls
    RunnerRegistry.get_or_create(), triggering replay_historical_events()
  - Store _is_resume flag in OpenHandsApp.__init__
  - Call ensure_runner() in _initialize_main_ui() after set_loaded_resources()
    and before _process_queued_inputs() (ordering constraint)

  BUG-002 (Medium): Rapid clicking in /history panel crashes the app
  - Wrap self.screen.get_selected_text() in on_mouse_up with try/except
  - KeyError from Toast widgets in transient lifecycle states is now caught
  - References upstream Textual issue #5646

  BUG-003 (Medium): --resume last throws ValueError instead of resuming
  - Add 3-line normalization at top of handle_resume_logic(): detect
    args.resume.lower() == last, set args.last = True, args.resume =
  - Case-insensitive: last, LAST, Last all work

  BUG-004 (Low): UnicodeDecodeError in containerized terminals
  - Documented as upstream Textual issue in .dev/KNOWN_ISSUES.md
  - No code change; includes reproduction steps and workaround

  BUG-005 (Low): RequestsDependencyWarning on stderr at launch
  - Add targeted warnings.filterwarnings() before third-party imports
  - Add E402/I001 per-file-ignore for entrypoint.py in pyproject.toml

  Test coverage:
  - tests/test_entrypoint.py: 5 new tests (--resume last normalization)
  - tests/tui/test_auto_copy.py: 1 new test (exception-path for BUG-002)
  - tests/tui/core/test_conversation_manager.py: 2 new tests (ensure_runner)

  Files changed: 3 source, 1 config, 1 doc, 3 test
  Lines: +86 -3
  All 1276 tests pass. ruff/pyright clean.

Development history: https://github.com/IronbellyOrg/Ironhands-CLI_dev/commit/e84a9c362d480adc629602e7607f3e17a9c63ba5

---
Co-authored-by: Ironbelly Ryan <ryan@ironbelly.com>


malhotra5:
@openhands /codereview-roasted

please consider separation of concerns, which components should be orchestrating the replay at all. check whether we are following existing patterns. post back the review as a github comment

malhotra5:
## 🔴 Taste Rating: Needs Improvement

**KEY INSIGHT:** Replay responsibility is scattered across 3 layers (App, Registry, Controller) when it should be concentrated in one controller following the existing message-based pattern.

---

### **[CRITICAL ISSUES]**

**[openhands_cli/tui/core/runner_registry.py, Lines 50-53] 🔧 Separation of Concerns:**
```python
# Replay historical events for newly created runners (resume / switch).
runner.replay_historical_events()
```
\`RunnerRegistry\` is a **dumb cache** — its job is to store and retrieve runners. Adding side-effect logic (replay) to \`get_or_create()\` violates single responsibility. The existing pattern has controllers orchestrate behavior, registries just store things.

**Suggestion:** Remove the replay call from \`RunnerRegistry\`. Have the caller (controller) trigger replay explicitly.

---

**[openhands_cli/tui/core/conversation_manager.py, Lines 297-306] 🔧 Pattern Violation:**
```python
def ensure_runner(self, conversation_id: uuid.UUID) -> None:
    """Eagerly create a runner for a conversation, triggering event replay."""
    self._runners.get_or_create(conversation_id)
```
The module docstring says ConversationManager is "intentionally small: it listens to... messages bubbling up... and delegates... to focused controllers." But \`ensure_runner()\` is an **imperative escape hatch** that bypasses the message architecture.

\`OpenHandsApp\` now calls \`conversation_manager.ensure_runner()\` directly instead of posting a message. This creates tight coupling and makes the flow invisible to the message-based debugging tools.

**Suggestion:** Either:
1. Create a \`ResumeConversation\` message that ConversationManager handles via a controller, OR
2. Make \`ConversationSwitchController\` the sole owner of replay orchestration (it already handles the switch case)

---

**[openhands_cli/tui/textual_app.py, Lines 460-470] 🔧 Layer Violation:**
```python
# BUG-001: For --resume sessions, eagerly create the runner so that
# replay_historical_events() fires before the user sends their first message.
# ORDERING CONSTRAINT: must be called after set_loaded_resources()...
if self._is_resume and self.conversation_state.conversation_id is not None:
    self.conversation_manager.ensure_runner(...)
```
The App layer now has knowledge of:
1. Runner creation timing (\`_is_resume\` flag)
2. Internal ordering constraints (after \`set_loaded_resources()\`, before \`_process_queued_inputs()\`)

This is exactly backwards — the **lower layers should handle their own initialization**. The App should just say "I'm resuming conversation X" and let the core layer figure out the details.

---

### **[IMPROVEMENT OPPORTUNITIES]**

**[openhands_cli/tui/core/conversation_runner.py, Line 278] ⚡ Simplification:**
```python
events = list(self.conversation.state.events)
```
Why copy to a list? If \`conversation.state.events\` is already iterable, passing it directly to \`replay_events()\` avoids memory allocation. Only materialize if you need \`len()\` — but you return \`len(events)\` after the fact, which could be \`sum(1 for _ in ...)\` or tracked during iteration.

---

**[Architectural] 🏗️ Replay Ownership Fragmentation:**
Replay is currently triggered from:
1. \`RunnerRegistry.get_or_create()\` (for all new runners)
2. \`OpenHandsApp._initialize_main_ui()\` (for resume via ensure_runner)
3. Implicitly via \`ConversationSwitchController._prepare_switch()\` → \`runners.get_or_create()\`

This is the "shotgun surgery" anti-pattern. If you need to change replay behavior, you touch 3+ files.

**Suggestion:** Concentrate replay in ONE place:
- **Option A**: Create a \`ReplayController\` that listens for \`ConversationActivated\` messages
- **Option B**: Make \`ConversationSwitchController\` handle both switch AND resume (they're conceptually the same: "make conversation X active")
- **Option C**: Use \`ConversationContainer.watch_conversation_id()\` to trigger replay when conversation_id changes (reactive pattern matches existing architecture)

---

### **[TESTING GAPS]**

**[tests/tui/core/test_conversation_runner_replay.py] 🧪 Mocks Aren't Tests:**
All 4 tests mock the conversation, visualizer, and events. There's no test that exercises the real \`replay_events()\` code path with actual \`MessageEvent\`, \`ActionEvent\`, \`ObservationEvent\` objects to verify:
1. User messages render with \`"> "\` prefix
2. Observations pair correctly with preceding actions
3. The scroll-to-end actually works

**Suggestion:** Add at least one integration test with real event types (can use fixtures from \`tests/tui/widgets/\` if they exist).

---

### **VERDICT:**
❌ **Needs rework** — The core functionality is sound but the architectural decomposition violates the existing message-based patterns. Concentrating replay ownership in a controller (probably \`ConversationSwitchController\` since it already handles the similar "make conversation active" flow) would make this much cleaner.

The BUG-002 (KeyError catch), BUG-003 (\`--resume last\` normalization), and BUG-005 (warnings filter) fixes are fine — minimal, targeted changes.

openhands_cli/tui/core/conversation_runner.py
```diff
@@ -266,6 +267,27 @@ def pause_runner_without_blocking(self):
...
+
+        Returns:
+            Count of replayed events, or 0 if already replayed or empty.
+        """
+        if self._historical_events_replayed:
+            return 0
+
+        self._historical_events_replayed = True
+
+        events = list(self.conversation.state.events)
```
enyst: Just to note, sorry for the quick passing, this is actually problematic since it’s O(n), and on long running conversations, this can reach funny numbers like **_60k events_**. Rendering them may probably bring a lot of machines/terminals to a halt. 😅 

That said, if this is the only place in the PR code, I don’t have a strong opinion on whether this PR needs to concern itself with it. I do think this is a major bug, but it’s pre-existing.

Please take a look at:
- https://github.com/OpenHands/software-agent-sdk/issues/1824
- https://github.com/OpenHands/OpenHands-CLI/issues/426

enyst: To note, this isn't an answer to my concern, is it?
It's an answer to the "roasted" top-level review comment, maybe it belongs on top-level too 😄

enyst:
(Totally on a side note, sorry: Rohit, you are using GPT-5.2 ? 👯 )

openhands_cli/tui/core/conversation_runner.py
ironbelly: I've added a commit addressing this, more details here https://github.com/IronbellyOrg/IronHands-CLI/pull/3

ironbelly:
ADDRESSED (from v0.02 adversarial review):
  1. Registry cache-before-replay ordering bug: The reviewer correctly
     identified that caching a runner before replay completes could leave a
     partially-initialized runner in the registry if replay throws. Fixed
     by reordering (T01.01) and adding regression tests (T01.02).
  2. Missing replay test coverage: Zero tests existed for replay_events().
     The reviewer's 8-scenario test matrix (T-1 through T-8) was adopted
     verbatim. These now guard rendering order, scroll behavior, and
     intentional side-effect omissions.
  3. Misleading ConversationManager docstring: The reviewer flagged that
     the docstring implied message-based dispatch was the exclusive
     coordination mechanism, when ensure_runner() and
     reload_visualizer_configuration() are synchronous direct calls. Fixed
     with qualified language.
  4. E402 lint violations: Pre-existing CI failure in entrypoint.py.
     Added noqa: E402 since the interleaving is intentional (BUG-005).

  PUSHED BACK (deferred or declined):
  1. Runtime thread-safety assertion in get_or_create(): The reviewer
     suggested adding `assert threading.current_thread() is
     threading.main_thread()`. We documented the constraint in the
     docstring (T01.03) but deferred the runtime check to a future release.
     Rationale: adding a hard assertion risks breaking test harnesses that
     run from non-main threads, and the docstring + test coverage provides
     sufficient protection for v0.03.
  2. Moving replay_historical_events into the factory: The adversarial
     panel considered moving replay into RunnerFactory.create() to make the
     invariant structural. The 4-expert consensus chose the registry guard
     approach because it preserves recovery flexibility — the caller can
     catch the exception and retry or fall back without the factory being
     opinionated about lifecycle. See Appendix B in tasklist.
  3. Snapshot environment parity: The CI snapshot failure cannot be
     reproduced locally. Rather than blindly regenerating snapshots in CI
     (which masks real regressions), we documented the environment delta
     and pinned the local enviro to the same version of python

> ## 🔴 Taste Rating: Needs Improvement
> 
> **KEY INSIGHT:** Replay responsibility is scattered across 3 layers (App, Registry, Controller) when it should be concentrated in one controller following the existing message-based pattern.

enyst:
@openhands Do a /codereview-roasted on this PR. Publish your review with the appropriate event on gh.

enyst (changes requested):
🔴 **Taste Rating: Needs Improvement**

**[CRITICAL ISSUES]**
- [openhands_cli/tui/widgets/richlog_visualizer.py, Lines 963–987] **State Tracking**: `load_older_events()` decrements `_loaded_start_index` by `mounted` (widget count). If any event doesn't render a widget (filtered MessageEvent, CondensationRequest, empty user message), `_loaded_start_index` stalls and PageUp keeps reprocessing the same slice forever. Decrement by the number of events consumed (e.g., `len(older_events)`), and keep `mounted` only for scroll offset, or track a separate `loaded_event_count`.
- [openhands_cli/tui/core/conversation_runner.py, Lines 301 & 353] **Idempotence**: replay guard only checks `_replayed_event_offset > 0` and sets it to `len(plan.tail_events)`. When the tail is empty (e.g., latest event is `Condensation`), offset stays 0 so replay can run repeatedly, duplicating banners/widgets. Use a boolean `replay_done` or set offset to `total_count` to mark replay even when tail is empty.

**[TESTING GAPS]**
- Add regression tests for (a) replay with empty tail after condensation being idempotent and (b) `load_older_events()` advancing `_loaded_start_index` even when older events produce no widgets.

**VERDICT:** ❌ **Needs rework**

**KEY INSIGHT:** You're using widget count as the history index, which breaks pagination and replay idempotence in edge cases.

enyst (commented):
This is a cool PR, thank you for this! Interesting workflow too ❤️

I say let’s let the agents battle it out 🫠

ironbelly:
> This is a cool PR, thank you for this! Interesting workflow too ❤️
> 
> I say let’s let the agents battle it out 🫠

Posted another update with those issues addressed.. Also lead to refactoring some of my skills and agents to better catch this kind of thing in the future, so thanks for that

ironbelly:
> 🔴 **Taste Rating: Needs Improvement**
> 
> **[CRITICAL ISSUES]**
> 
> * [openhands_cli/tui/widgets/richlog_visualizer.py, Lines 963–987] **State Tracking**: `load_older_events()` decrements `_loaded_start_index` by `mounted` (widget count). If any event doesn't render a widget (filtered MessageEvent, CondensationRequest, empty user message), `_loaded_start_index` stalls and PageUp keeps reprocessing the same slice forever. Decrement by the number of events consumed (e.g., `len(older_events)`), and keep `mounted` only for scroll offset, or track a separate `loaded_event_count`.
> * [openhands_cli/tui/core/conversation_runner.py, Lines 301 & 353] **Idempotence**: replay guard only checks `_replayed_event_offset > 0` and sets it to `len(plan.tail_events)`. When the tail is empty (e.g., latest event is `Condensation`), offset stays 0 so replay can run repeatedly, duplicating banners/widgets. Use a boolean `replay_done` or set offset to `total_count` to mark replay even when tail is empty.
> 
> **[TESTING GAPS]**
> 
> * Add regression tests for (a) replay with empty tail after condensation being idempotent and (b) `load_older_events()` advancing `_loaded_start_index` even when older events produce no widgets.
> 
> **VERDICT:** ❌ **Needs rework**
> 
> **KEY INSIGHT:** You're using widget count as the history index, which breaks pagination and replay idempotence in edge cases.

This has all been addressed in the recent update

enyst:
@openhands Read all top level comments here, and understand this PR in depth.

Another AI agent has previously posted a REQUEST_CHANGES review; and the PR author and their agents have fixed the PR since then. Let's see if the fixed PR is fine. Do a /codereview-roasted on it and publish your review. You can use the suitable `event` in the PR review via gh api.

Think deeply and publish a review (you are not allowed to post a comment, but publish a review, you are allowed to publish a review with the gh api and with an event)

enyst (dismissed):
🔴 Taste Rating: Needs Improvement

**[CRITICAL ISSUES]**
- [openhands_cli/tui/widgets/richlog_visualizer.py, Lines 1040–1064] **Condensation banner disappears after full load**: `_update_banner_for_loaded_state()` removes `_banner_widget` when `_loaded_start_index` reaches 0, even if `_has_condensation` is true. After the user pages up to load all post‑condensation events, the summary banner is removed and there is no indication of condensed prior context. This breaks the goal of preserving condensed history. Keep the banner when `has_condensation` is true, and update its text based on `condensed_count` even when `_loaded_start_index` is 0.

**[TESTING GAPS]**
- Add a regression test that loads older events until `_loaded_start_index == 0` in a condensation scenario and asserts the summary banner remains mounted (or updated) with the condensed‑count text. The current tests only cover non‑condensation load paths.

**VERDICT:** ❌ Needs rework
**KEY INSIGHT:** The summary banner is treated as a “load‑more” indicator, but for condensation it’s the only representation of hidden history and must persist.

ironbelly:
I see I have some cleaning up to do.  Feeling guilty for the time and resources I'm taking to babysit this PR with me.  Give me a couple days to review my process and implement the same types of checks on my side as you have on your side

ironbelly:
@openhands-agent 

# Objective: 
 - To identify faults and weaknesses in my workflow in order to strengthen my process and prevent similar issues from occurring in the future

# Task: 
- In regards to the Critical issue you listed.  Read through our conversation history and previous feedback as well as file structure and comments and determine if there we're clear indicators or instructions requiring the Condensation Banner to persist when has_condensation is true, and update its text based on condensed_count even when _loaded_start_index is 0.  

# Context: 
`[openhands_cli/tui/widgets/richlog_visualizer.py, Lines 1040–1064] Condensation banner disappears after full load: _update_banner_for_loaded_state() removes _banner_widget when _loaded_start_index reaches 0, even if _has_condensation is true. After the user pages up to load all post‑condensation events, the summary banner is removed and there is no indication of condensed prior context. This breaks the goal of preserving condensed history. Keep the banner when has_condensation is true, and update its text based on condensed_count even when _loaded_start_index is 0.`

ironbelly:
@openhands-agent 

# Objective: 
 - To identify faults and weaknesses in my workflow in order to strengthen my process and prevent similar issues from occurring in the future

# Task: 
- In regards to the Critical issue you listed.  Read through our conversation history and previous feedback as well as file structure and comments and determine if there we're clear indicators or instructions requiring the Condensation Banner to persist when has_condensation is true, and update its text based on condensed_count even when _loaded_start_index is 0.  

# Context: 
`[openhands_cli/tui/widgets/richlog_visualizer.py, Lines 1040–1064] Condensation banner disappears after full load: _update_banner_for_loaded_state() removes _banner_widget when _loaded_start_index reaches 0, even if _has_condensation is true. After the user pages up to load all post‑condensation events, the summary banner is removed and there is no indication of condensed prior context. This breaks the goal of preserving condensed history. Keep the banner when has_condensation is true, and update its text based on condensed_count even when _loaded_start_index is 0.`

enyst:
Thank you for this PR, and I'm really sorry if my agent was too obnoxious.

Just for clarity, IMHO it's perfectly possible to limit events rendered at restore in more than one way: 
- maybe up to 20 (or hardcoded value)
- maybe up to last condensation
- maybe by paging - as I think this PR did

My only concern here was that we cannot, IMHO, render ALL history events. The amount can be large (unbounded practically, really) and have the bad habit to slow down the machine maybe as soon as 1-2k

.gitignore
enyst, line 1: Is there a reason why this is deleted?

pyproject.toml
```diff
@@ -91,6 +91,9 @@ extend-select = ["B006", "B008", "B039", "RUF012"]
 [tool.ruff.lint.per-file-ignores]
 # Test files often have unused arguments (fixtures, mocks, interface implementations)
 "tests/**/*.py" = ["ARG"]
+# BUG-005: entrypoint places a warnings.filterwarnings() call between stdlib and
+# third-party imports to suppress RequestsDependencyWarning before it fires.
+"openhands_cli/entrypoint.py" = ["E402", "I001"]
```
enyst, line 96: Is this necessary?

tests/tui/core/test_runner_registry.py
```diff
@@ -0,0 +1,85 @@
...
+        runner_a.conversation = MagicMock()
+        runner_b = MagicMock()
+        runner_b.conversation = MagicMock()
+        mock_factory.create.side_effect = [runner_a, runner_b]
+
+        registry.get_or_create(cid_a)
+        registry.get_or_create(cid_b)
+
+        runner_a.replay_historical_events.assert_called_once()
+        runner_b.replay_historical_events.assert_called_once()
```
enyst, line 65: A lot of tests is nice! Though I wonder if we need them so fine-grained that they only test calls on mocks. We kinda try lately to keep the tests a little more realistic, less mocks, more integration-like testing of behavior, rather than calls per se.

The main point is that it seems more useful, and with LLMs happily generating everything under the sun, it’s kinda difficult for humans to keep up. 🫠

openhands_cli/entrypoint.py
```diff
@@ -47,6 +53,12 @@ def handle_resume_logic(args: argparse.Namespace) -> str | None:
     Returns:
         Conversation ID to resume, or None if it should show conversation list or exit
     """
+    # BUG-003: Normalize `--resume last` (case-insensitive) to the --last code path.
+    # Users intuitively type `--resume last` instead of `--resume --last`.
+    if args.resume and args.resume.lower() == "last":
+        args.last = True
```
enyst, line 59: Nit: Is this tiny detail necessary?

openhands_cli/entrypoint.py
```diff
@@ -11,17 +11,23 @@
 import warnings
 from pathlib import Path
 
-from dotenv import load_dotenv
-from rich.console import Console
+# BUG-005: Suppress RequestsDependencyWarning before third-party imports.
+# The `requests` library (via urllib3) emits this warning at import time when
+# charset_normalizer is used instead of chardet. Filter must be active before
+# any transitive import of `requests` (e.g., via openhands_cli.stores).
+warnings.filterwarnings("ignore", message=r"urllib3.*or chardet.*charset_normalizer")
```
enyst, line 18: Could you maybe elaborate on what happened here?

tests/tui/core/test_nfr5_guard.py
```diff
@@ -0,0 +1,34 @@
+"""NFR-5 guard: forbid list(state.events) materialization in replay-related files.
+
+This test enforces the v0.04 non-functional requirement that replay code must
+never eagerly materialise the full event list via ``list(state.events)`` or
+``list(self.conversation.state.events)``.  The windowed/lazy replay cascade
+should be the only access path.
```
enyst, line 6: This is a funny and interesting test! I think maybe we shouldn’t need it, but I can feel why your LLM council may have wanted it 😅

Although… this is the kind of stuff we may want in CI someday 🤔 but I’m not sure yet under what form

tests/tui/core/test_integration_load_older.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # Initial replay should use window=2
+        count = runner.replay_historical_events()
+        assert count == 2
+
+        # Visualizer context should be set with all 5 events
+        assert vis._loaded_start_index == 3  # 5 - 2 = 3
+        assert len(vis._all_events) == 5
+
+        # Now simulate a banner widget for the prepend path
```
enyst, line 100: This seems a little suspicious, why simulate? 😅

.serena/project.yml
enyst, line 1: I think maybe this file wasn’t intended? 😀

ironbelly: Not really, helpful if you all use Serena and want to pool and share memory across sessions.. Sorry about that

tests/tui/core/test_integration_load_older.py
ironbelly: I've noticed Claude doing this a LOT lately.. Instead of putting actual functoinality in or proper tests it will simulate and mock everywhere it can.  Did a deep dive on this with an agent today and got this response

LLMs are strongly biased toward producing outputs that look complete and
  well-structured over outputs that are actually complete. The roadmap has
  10 phases, milestones, timelines, success criteria, a validation matrix —
  it looks thorough. The dispatch wiring is a single unglamorous integration
   line that doesn't merit its own phase heading. LLMs gravitate toward the
  architecturally interesting (the executor loop, the gate system, the
  convergence loop) and away from the mundane connective tissue.

tests/tui/core/test_nfr5_guard.py
ironbelly: I'm sure the council won't take it personally

.gitignore
ironbelly: shouldn't have been deleted, I just added it to my own gitignore so I could get more fine tuned and not pollute upstream with my .gitignore rules.. Mainly I load my .dev directory where all of the artifacts live as a gitmodule and wanted to add .gitmodules and .dev to .gitignore

tests/tui/core/test_runner_registry.py
ironbelly: See my other comment below.. I have been loosing my mind at Claude this week as it has been converting real tests into mocks without my permission or instructions in order to produce 'valuable looking' results without having to spend the tokens.  Giving the impression of value is far more important for it than delivering real value.

openhands_cli/entrypoint.py
ironbelly: Probably not, just a tiny QOL thing from someone that always types --resume last :)

openhands_cli/entrypoint.py
ironbelly: This might be related to running Openhands in the specific environment I have here. 
All of our dev environments here run in code-server containers managed by Coder.  So headless VS Code terminals, in Chrome, through Coder.  In this environment there were warnings like those above and this was one oft he fixes

tests/tui/core/test_integration_load_older.py
enyst: Sonnet 3.7 had a very strong tendency to this behavior: create placeholders, “simulate”, assert 1=1 🫠

Other Claudes do better these days, but it still happens. In other words, it will cheat to make tests pass, a bit visibly; probably reward hacking. I don’t think that response covers it 😅

enyst: This is part of the reason why we are trying to move towards real-like tests, integration tests, snapshot tests, less mocking ; the humans cannot figure out easily if it tests for real or not

openhands_cli/entrypoint.py
enyst: Fair enough, thank you 🙏

PR #581, opened 2026-03-08
tui: add tab navigation hint to footers (#512)

### LINKED ISSUE #512: [CLI] Using "Tab" to navigate in CLI is atypical
We received feedback for the settings page of the CLI, where the user did not realize they could use the tab key to navigate the settings field. This probably applies to other parts of the CLI as well.

> was very confusing to me, I thought it was broken 
maybe we just need a note to use tab? I'm not used to tabbing around inside of a CLI

ixchio:
# Description
Fixes #512.

As per the user feedback, navigating fields inside the CLI modals and screens exclusively with the `Tab` key was atypical and unintuitive, leaving users to think the application was broken. 

To improve discoverability:
1. Added the `Tab` key binding globally to the main Application via `OpenHandsApp`.
2. Modified the `SettingsScreen` modal to explicitly include a `Footer` and added the `Tab` key to its bindings list as well.

## Evidence of Changes
- Ran `make lint` and the changes are formatting-compliant.
- Ran `make test` for regression tests.
- Re-generated visual SVGs via `make test-snapshots --snapshot-update` to capture the new footer text logic showing "Tab Navigate".

I hope this helps new users get to their setup much quicker! Let me know if there's anything else you need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,20 +154,54 @@ MIT License - see [LICENSE](LICENSE) for details.
 
 <hr>
 <div align="center">
-  <sub><strong>Trusted by engineers at:</sub></strong>
-  </br> 
-  </br>
-  <img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137f6974b71a1a4a932f82_TikTok_logo.svg" alt="TikTok" height="30">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
-  <img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137f523b08f91a5aa905b9_Vmware.svg" alt="VMware" height="30">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
-  <img src="https://cdn.prod.website-files.com/68ff4058b35616cdd47d5b59/69137f2cb537758796a9dba1_Roche_Logo.svg" alt="Roche" height="30">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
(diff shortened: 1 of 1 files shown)


ixchio:
> Hey @ixchio , thank you for the PR!
> 
> It seems, from the snapshot differences, that the Save button is not in the center anymore? And the footer text is almost "cut off" on top, it seems - could you maybe look into what happened there?

Thanks for pointing that out! I didn't notice the layout shift in the snapshots. I'll take a look at the Save button alignment and the footer clipping issue and push a fix shortly.

ixchio:
## Fix Applied

@enyst Thank you for the review feedback! I've identified and fixed the layout issue.

### Problem
The `Footer()` widget was incorrectly placed **inside** the `settings_container` Container, which caused:
1. The Save button to shift from center alignment
2. The footer text to appear "cut off" within the modal content area

### Solution
Moved the `Footer()` yield statement **outside** the `with Container(id="settings_container"):` block, so it now renders at the `ModalScreen` level instead of inside the modal content.

### Changes Made
- `openhands_cli/tui/modals/settings/settings_screen.py`: Moved `yield Footer()` outside the container context
- Updated snapshots to reflect the corrected layout

### Evidence
The snapshots now show:
- Save and Cancel buttons properly centered in the modal
- Footer with "tab Navigate esc Cancel" correctly positioned at the bottom of the screen

Commit: 052cbc2

ixchio:
Hey @enyst  just checking in on this one! I've addressed all the layout feedback (Footer moved outside the container, snapshots updated). Been open for 2 weeks now, would love to get this across the finish line. 
Does the latest look good to you? 🙏

PR #612, opened 2026-03-24
fix: support AWS IAM credentials for Bedrock models without LLM_API_KEY

### LINKED ISSUE #611: [Bug]: Can't authenticate with AWS IAM credentials with CLI
### Is there an existing issue for the same bug?

- [x] I have searched existing issues and this is not a duplicate.

### Bug Description

I can't authenticate with AWS IAM credentials with CLI to run on bedrock. It works in the SDK, but when I run `openhands --override-with-envs` it says it wants an `LLM_API_KEY`. I tried talking with Claude and ChatGPT and Gemini and tried all the stuff they suggested like using a fake API key. 

I also tried setting stuff in the GUI that pops up under advanced but no dice. 

As I said, authenticating with the SDK works, but the CLI doesn't seem willing to fall back to the AWS IAM credentials.

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Steps to Reproduce

_No response_

### OpenHands CLI Installation Method

CLI (uv tool install)

### If you selected "Other", please specify

_No response_

### OpenHands CLI Version

v1.13.1

### Version Confirmation

- [x] I have confirmed this bug exists on the LATEST version of OpenHands CLI

### Model Name

bedrock/arn:aws-us-gov:bedrock:us-gov-west-1:240894443918:inference-profile/us-gov.anthropic.claude-sonnet-4-5-20250929-v1:0

### Operating System

MacOS

### Terminal / Shell

_No response_

### Logs and Error Messages

_No response_

### Screenshots and Additional Context

_No response_

ixchio:
Hey! 👋

So this PR fixes the issue where you couldn't use AWS Bedrock with the CLI using IAM credentials.

## What was happening?

When running `openhands --override-with-envs` with a Bedrock model, the CLI would complain about missing `LLM_API_KEY` - but Bedrock doesn't use API keys! It uses AWS IAM credentials instead.

## The fix

Now the CLI is smart enough to detect when you're using a Bedrock (or SageMaker) model and won't ask for an API key. It'll use your AWS credentials instead, which can come from:

- Environment variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_REGION_NAME`)
- Your `~/.aws/credentials` file
- IAM roles if you're on EC2/ECS/Lambda

## How to use it

Just set your model and go:

```bash
export LLM_MODEL="bedrock/anthropic.claude-3-sonnet"
openhands --override-with-envs
```

Or with the full ARN like in the issue:

```bash
export LLM_MODEL="bedrock/arn:aws-us-gov:bedrock:us-gov-west-1:123456:inference-profile/us-gov.anthropic.claude-sonnet-4-5-20250929-v1:0"
openhands --override-with-envs
```

No need to set a fake API key or anything like that anymore!

## Testing

- All 1275 existing tests pass ✅
- Added 15 new tests for the AWS auth path
- Manually verified with the exact model from the issue

Fixes #611


enyst (commented):
Thank you for the PR! 

I'm a bit curious, do we need to use `--override-with-envs` for AWS at all?

I believe the SDK has now LLM class attributes for Bedrock, maybe we could use those?

ixchio:
> Thank you for the PR!
> 
> I'm a bit curious, do we need to use `--override-with-envs` for AWS at all?
> 
> I believe the SDK has now LLM class attributes for Bedrock, maybe we could use those?

Thanks @enyst, you're right! I checked and the SDK (openhands-sdk v1.16.1) already has native AWS/Bedrock support built into the LLM class:

    8 dedicated AWS fields: aws_access_key_id, aws_secret_access_key, aws_session_token, aws_region_name, aws_profile_name, aws_role_name, aws_session_name, aws_bedrock_runtime_endpoint
    LLM.load_from_env(prefix="LLM_") auto-reads them from LLM_AWS_* env vars (e.g., LLM_AWS_ACCESS_KEY_ID, LLM_AWS_REGION_NAME)
    The LLM constructor already works fine without api_key for bedrock models

So this PR could be simplified significantly. The only thing the CLI really needs to fix is:

    Don't require LLM_API_KEY when the model has a bedrock/ or sagemaker/ prefix
    Let the SDK handle the rest — users can set LLM_AWS_ACCESS_KEY_ID, LLM_AWS_REGION_NAME, etc., and LLM.load_from_env() picks them up, or boto3's default credential chain handles ~/.aws/credentials / IAM roles automatically

I'll simplify the PR to remove the custom AWS credential plumbing from LLMEnvOverrides and just lean on the SDK's existing support. Want me to go that route?

ixchio:
Conflict’s fixed and I simplified the AWS auth path to lean on the SDK’s LLM.load_from_env() / LLM_AWS_* support instead of custom CLI credential plumbing.

enyst:
Thank you!

enyst:
@openhands Do a /codereview on this PR. You don't have bedrock creds, but you can investigate docs and understand the desired behavior. Then post your feedback as a review with gh api, with approved or comment or request changes as you think appropriate.

PR #636, opened 2026-04-02
feat: ensure backward compatibility for persisted conversations with DelegateTool

### LINKED ISSUE #635: Ensure backward compatibility for persisted conversations when switching to TaskToolSet
## Summary

Ensure backward compatibility for persisted conversations when migrating from `DelegateTool` to `TaskToolSet`.

## Background

When replacing `DelegateTool` with `TaskToolSet`, we need to handle persisted conversations that may contain `DelegateTool` events. Our current system allows us to add tools, but not to replace or delete them without consideration.

## Proposed Solution

Implement conditional tool loading based on conversation history:

1. **If the persisted conversation has `DelegateTool` events**: Add `DelegateTool` to the tool set (not `TaskToolSet`)
2. **If no `DelegateTool` in the conversation**: Add `TaskToolSet`

This approach allows:
- Existing conversations to continue working with their original tooling
- New conversations to use the improved `TaskToolSet`
- Gradual deprecation of `DelegateTool`

## Related Work

- Similar challenge addressed in software-agent-sdk PR #2064, which removes `undo` from `file_editor` tool
- The strategy there: always read those events; map them to something innocuous (in that case, "view")

## Acceptance Criteria

- [ ] Detect if a persisted conversation contains `DelegateTool` events
- [ ] Conditionally load the appropriate tool based on conversation history
- [ ] Add tests for both scenarios (legacy and new conversations)
- [ ] Document the deprecation timeline for `DelegateTool`

---
*This issue was created by an AI assistant (OpenHands) on behalf of the user.*

VascoSch92:
## Summary

This PR implements backward compatibility for persisted conversations when migrating from `DelegateTool` to `TaskToolSet`, as described in issue #635.

**Note**: the test cases are made from old conversation I had in my archive.

### Core Implementation

1. **New default tools**: Changed `get_default_cli_tools()` to use `TaskToolSet` by default instead of `DelegateTool` for new conversations.

2. **Backward compatibility detection**: Added `conversation_has_delegate_tool_events(conversation_id)` function that scans conversation event files for any `ActionEvent` with `tool_name="delegate"`.

3. **Conditional tool loading**: Updated `_resolve_tools()` in `AgentStore` to:
   - Use persisted tools from `base_state.json` if available (existing behavior)
   - If no persisted tools, check for `DelegateTool` events in conversation history
   - Use `DelegateTool` for conversations with delegate events (backward compatibility)
   - Use `TaskToolSet` for new conversations or conversations without delegate events

### Test Coverage

- Added comprehensive tests for `conversation_has_delegate_tool_events()`:
  - Non-existent conversations
  - Empty events directory
  - Events without delegate tool usage
  - Events with delegate tool usage
  - Graceful handling of invalid JSON files

- Added tests for `_resolve_tools()` backward compatibility:
  - New conversations use `TaskToolSet`
  - Conversations with delegate events use `DelegateTool`
  - Conversations without delegate events use `TaskToolSet`
  - Persisted tools take precedence over detection
  - Non-existent conversations fall back to `TaskToolSet`

- Updated existing tests to expect `task_tool_set` for new conversations

## How It Works

```
New conversation (no session_id):
  └── Use TaskToolSet (default for new conversations)

Resumed conversation (with session_id):
  ├── Has base_state.json with tools?
  │   └── Yes → Use persisted tools
  │   └── No → Check events for delegate usage
  │       ├── Has delegate events?
  │       │   └── Yes → Use DelegateTool (backward compatibility)
  │       │   └── No → Use TaskToolSet
```

## Verification

Commands run:
- `make lint` ✅
- `make test` ✅ (1289 tests passed)

Fixes #635

---
*This PR was created by an AI assistant (OpenHands) on behalf of the user.*

@VascoSch92 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/089169e9-9c16-43c9-95ed-966c648253c6)

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@openhands/backward-compat-delegate-tool-635
```


openhands_cli/utils.py
```diff
@@ -160,13 +163,56 @@ def get_llm_metadata(
...
+    Returns:
+        True if the conversation contains DelegateTool events, False otherwise
+    """
+    conversations_dir = get_conversations_dir()
+    events_dir = Path(conversations_dir) / conversation_id / "events"
+
+    if not events_dir.exists():
+        return False
+
+    for event_file in events_dir.glob("event-*.json"):
```
xingyaoww: Can we just load the SystemPromptEvent? I remember that has a `.tools` field which should be more efficient than the traversal here

openhands_cli/utils.py
```diff
@@ -160,13 +163,56 @@ def get_llm_metadata(
     return metadata
 
 
-def get_default_cli_tools() -> list[Tool]:
-    """Get the default tool specifications for CLI mode (browser disabled)."""
+def conversation_has_delegate_tool_events(conversation_id: str) -> bool:
```
xingyaoww: nit: maybe move this to a different file, eg `deprecated_utils`?

enyst (commented):
Thank you for this, we could do something like this… 🤔 

I have two little questions

1. The SDK does the restoring of a conversation, and it has the check for tools that have been removed/added. It feels a bit that maybe this could be there?
2. How do we think about [this solution](https://github.com/OpenHands/software-agent-sdk/issues/2667#issuecomment-4178506313), is there something… problematic with it?

openhands_cli/utils.py
VascoSch92: Done.

openhands_cli/utils.py
VascoSch92: Yes. It is true. Much better. Changed.

VascoSch92:
> Thank you for this, we could do something like this… 🤔
> 
> I have two little questions
> 
>     1. The SDK does the restoring of a conversation, and it has the check for tools that have been removed/added. It feels a bit that maybe this could be there?
> 
>     2. How do we think about [this solution](https://github.com/OpenHands/software-agent-sdk/issues/2667#issuecomment-4178506313), is there something… problematic with it?

1. I also think the logic should reside in the SDK and/or the agent-server. My understanding is that this is a short-term hotfix to maintain backward compatibility, as a proper solution will likely take some time to implement.

2. No problem with that at all. Let’s continue the discussion in the issue!

xingyaoww (approved):
Thank you

PR #658, opened 2026-04-13
fix(settings): reset max_input_tokens on model change

cbagwell:
When switching LLM models in the settings screen, max_input_tokens was carried over from the previous model's configuration. For example, switching from gemini-2.5-pro (1M context) to gpt-5.3-codex would incorrectly keep the 1M limit.

When a model change is detected in save_settings(), max_input_tokens is set to None so the SDK's LLM.model_post_init() auto-looks up the correct value from LiteLLM metadata. When the model is unchanged, the existing value is preserved.

The validate_max_tokens field validator already ensures data.max_tokens is int | None, so the previous inline int() conversion was redundant.


cbagwell:
This seemed simplest approach to solve immediate problem seen when using TUI to change model to one with smaller context limits.  Depending on how future "/model" feature is added, this logic may not be needed any more or look a lot different.

openhands_cli/tui/modals/settings/utils.py
```diff
@@ -198,9 +206,7 @@ def save_settings(
             timeout=int(data.timeout)
             if isinstance(data.timeout, str)
             else data.timeout,
-            max_input_tokens=int(data.max_tokens)
-            if isinstance(data.max_tokens, str)
```
enyst, line 202: For some reason, it seems this could be a string?

enyst (approved):
Thank you!

openhands_cli/tui/modals/settings/utils.py
cbagwell: I think it is a pyright specific narrowing issue but I reverted logic back to original isinstance() to make it happy again.  Sorry, I mistakenly assumed pyright was ran locally as part of "make test" and/or git pre-commit hook.

PR #663, opened 2026-04-14
makefile: avoid local .openhands writes during make test

llamantino:

---

## Why

Running make test locally can write into the local `~/.openhands` directory.

## Summary

- run `make test` with a temporary `HOME` to avoid local `~/.openhands` writes

## How to Test

Run `make test` and verify it does not create or write to the local `~/.openhands` directory.

## Type

- [X] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,17 +70,17 @@ install-dev: check-uv-version
 
 test: check-uv-version
 	@$(ECHO) "$(YELLOW)Run tests...$(RESET)"
-	uv run pytest --ignore=tests/snapshots
+	uv run env HOME="$$(mktemp -d)" pytest --ignore=tests/snapshots
 	@$(ECHO) "$(GREEN)Tests completed.$(RESET)"
 
 test-snapshots: check-uv-version
 	@$(ECHO) "$(YELLOW)Run snapshots tests...$(RESET)"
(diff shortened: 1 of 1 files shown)


enyst (approved):
🟢 **Good taste** - Elegant, simple solution.

Verified locally:
- On `main`: `HOME=$(mktemp -d) make test` completed successfully and created `TEST_HOME/.openhands/...`.
- On this PR: the same smoke run completed with `1308 passed in 59.01s` and `TEST_HOME/.openhands` was not created.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟢 LOW
This is a narrow Makefile-only change that fixes a real local side effect without touching application runtime behavior. The approach is pragmatic: it isolates `HOME` only for the pytest process, so the fix lands at the boundary where the unwanted writes happen.

**VERDICT:**
✅ **Worth merging**: Core logic is sound, and local verification matches the intended behavior.

**KEY INSIGHT:**
Scoping `HOME` at the test command is the simplest place to stop test-side persistence without changing application code.

_This review was created by an AI assistant (OpenHands) on behalf of the user._

PR #666, opened 2026-04-15
cli: replace prompt_toolkit with rich for styled output

llamantino:

---

## Why

`prompt_toolkit` was only being used for styled terminal output in a few places. The project already uses `rich` for console output, so keeping both adds unnecessary dependency and import overhead.

## Summary

- replace `prompt_toolkit` output formatting with `rich` console output
- remove `prompt_toolkit` as a dependency
- document `uv` 0.11.6 as the minimum supported version for consistent lockfile behavior

## How to Test

- use `uv` `0.11.6` or newer
- `make install-dev`
- `uv run pytest tests/test_gui_launcher.py`
- `uv lock` on this branch and verify `uv.lock` does not change
- `uvx --from uv==0.11.6 uv lock` and verify `uv.lock` still does not change

### Additional verification performed

- On `origin/main`, running `uv lock` with newer uv rewrites the old `exclude-newer` header to the current `7 days ago` cutoff and adds `exclude-newer-span = "P7D"`
- In a temporary repro project with `exclude-newer = "3 weeks"` and `idna`, re-locking later with `uv lock --upgrade-package idna` moved the stored cutoff forward and upgraded `idna` as expected

## Type

- [X] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The `uv.lock` timestamp here is expected to reflect the relative cutoff (`now - 7 days`), not the current day of the `uv lock` command.
- Minimum supported `uv` version is `0.11.6`; older versions can serialize `uv.lock` differently around relative `exclude-newer`.
- Contributors should keep `uv` updated for security fixes; this PR documents the minimum version, but does not itself upgrade the tool in local/CI environments.

_This PR description was updated by an AI assistant (OpenHands) on behalf of the user._

### DIFF
--- a/openhands_cli/gui_launcher.py
+++ b/openhands_cli/gui_launcher.py
@@ -6,23 +6,19 @@ import subprocess
 import sys
 from pathlib import Path
 
-from prompt_toolkit import print_formatted_text
-from prompt_toolkit.formatted_text import HTML
+from rich.console import Console
+from rich.markup import escape
 
 from openhands_cli.locations import get_persistence_dir
 
 
+console = Console(highlight=False, soft_wrap=True)
+
+
 def _format_docker_command_for_logging(cmd: list[str]) -> str:
-    """Format a Docker command for logging with grey color.
-
-    Args:
-        cmd (list[str]): The Docker command as a list of strings
-
-    Returns:
-        str: The formatted command string in grey HTML color
-    """
+    """Format a Docker command for logging."""
     cmd_str = " ".join(cmd)
-    return f"<grey>Running Docker command: {cmd_str}</grey>"
+    return f"Running Docker command: {cmd_str}"
 
 
 def check_docker_requirements() -> bool:
@@ -33,13 +29,15 @@ def check_docker_requirements() -> bool:
     """
     # Check if Docker is installed
     if not shutil.which("docker"):
-        print_formatted_text(
-            HTML("<ansired>❌ Docker is not installed or not in PATH.</ansired>")
+        console.print(
+            "❌ Docker is not installed or not in PATH.",
+            style="red",
+            markup=False,
         )
-        print_formatted_text(
-            HTML(
-                "<grey>Please install Docker first: https://docs.docker.com/get-docker/</grey>"
-            )
+        console.print(
+            "Please install Docker first: https://docs.docker.com/get-docker/",
+            style="grey50",
+            markup=False,
         )
         return False
 
@@ -49,18 +47,24 @@ def check_docker_requirements() -> bool:
             ["docker", "info"], capture_output=True, text=True, timeout=10
         )
         if result.returncode != 0:
-            print_formatted_text(
-                HTML("<ansired>❌ Docker daemon is not running.</ansired>")
+            console.print(
+                "❌ Docker daemon is not running.",
+                style="red",
+                markup=False,
             )
-            print_formatted_text(
-                HTML("<grey>Please start Docker and try again.</grey>")
+            console.print(
+                "Please start Docker and try again.",
+                style="grey50",
+                markup=False,
             )
             return False
     except (subprocess.TimeoutExpired, subprocess.SubprocessError) as e:
-        print_formatted_text(
-            HTML("<ansired>❌ Failed to check Docker status.</ansired>")
+        console.print(
+            "❌ Failed to check Docker status.",
+            style="red",
+            markup=False,
         )
-        print_formatted_text(HTML(f"<grey>Error: {e}</grey>"))
+        console.print(f"Error: {e}", style="grey50", markup=False)
         return False
 
     return True
@@ -92,10 +96,8 @@ def launch_gui_server(mount_cwd: bool = False, gpu: bool = False) -> None:
         gpu: If True, enable GPU support by mounting all GPUs into the
             container via nvidia-docker.
     """
-    print_formatted_text(
-        HTML("<ansiblue>🚀 Launching OpenHands GUI server...</ansiblue>")
-    )
-    print_formatted_text("")
+    console.print("🚀 Launching OpenHands GUI server...", style="blue", markup=False)
+    console.print()
 
     # Check Docker requirements
     if not check_docker_requirements():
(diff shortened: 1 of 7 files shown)


openhands_cli/gui_launcher.py
```diff
@@ -156,7 +160,8 @@ def launch_gui_server(mount_cwd: bool = False, gpu: bool = False) -> None:
     # Add current working directory mount if requested
     if mount_cwd:
         cwd = Path.cwd()
-        # Following the documentation at https://docs.all-hands.dev/usage/runtimes/docker#connecting-to-your-filesystem
+        # Following the documentation at
+        # https://docs.all-hands.dev/usage/runtimes/docker#connecting-to-your-filesystem
```
enyst:
```suggestion
        # https://docs.openhands.dev/usage/runtimes/docker#connecting-to-your-filesystem
```

Not an issue of this PR of course, just a tiny detail I happened to notice in passing, we could maybe just fix it

enyst:
@openhands /codereview

Post your feedback as a review using gh api with event.

enyst (changes requested):
🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [uv.lock:1665-1711 / pyproject.toml:25-28] **Incomplete dependency removal**: `prompt-toolkit` was removed from `pyproject.toml`, but `uv.lock` still records it as a direct dependency of `openhands`. This repo installs via `uv sync`, so the locked environment will still pull the package in. Right now the PR says the dependency is removed, but the install surface still says otherwise.
- [openhands-cli.spec:46-49, 89-91] **Binary build risk**: the PyInstaller spec still hard-codes `prompt_toolkit` in `hiddenimports` / `excludes`. If this dependency is really gone, `collect_submodules('prompt_toolkit')` is exactly the kind of stale packaging hook that breaks the frozen binary path.

**[TESTING GAPS]**
- [PR description] **No evidence**: `How to Test` and `Video/Screenshots` are still empty. For a dependency/packaging change, please include the exact end-to-end commands and output used to verify the real code path, and ideally build/binary verification as well. Unit-test edits alone are not enough proof here.

**VERDICT:**
❌ **Needs rework**: the runtime code change is straightforward, but the dependency removal is only partial across lock/install/build surfaces.

**KEY INSIGHT:**
The Rich migration itself is simple; the risky part is that `prompt_toolkit` still exists in the repo's packaging metadata, so users and release artifacts can still depend on it even after this PR merges.

_This review was created by an AI assistant (OpenHands) on behalf of the user._

enyst:
@openhands /codereview

Post your feedback as a review using gh api with event.

enyst (changes requested):
🟡 **Acceptable** - The code change itself is small and sensible, but there are still a couple of things that need to be cleaned up before this is ready.

**[CRITICAL ISSUES]**
- [PR description] **No Evidence**: The PR body still leaves `How to Test` and `Video/Screenshots` empty. For a user-visible CLI change, `"a human has tested these changes"` is not enough. Add the exact end-to-end command(s) you ran and the resulting output that proves the rich-styled paths still behave correctly in a real run.

**[IMPROVEMENT OPPORTUNITIES]**
- [uv.lock, lines 4-5] **Scope creep / lockfile policy drift**: This PR is about removing `prompt_toolkit`, but the lockfile also changes the `exclude-newer` metadata from a fixed timestamp to an older timestamp plus `exclude-newer-span = "P7D"`. That is a repo-wide dependency resolution policy change, not part of the feature. Keep the lockfile metadata stable unless that policy change is intentional and explained separately.

**VERDICT:**
❌ **Needs rework**: The implementation direction is fine, but the PR still lacks the proof that it works and includes an unrelated lockfile policy change.

**KEY INSIGHT:**
The core refactor is simple; the remaining risk is not the code path itself, but missing runtime evidence and unnecessary lockfile churn.

_This review was generated by an AI assistant (OpenHands) on behalf of the user._

uv.lock
```diff
@@ -3,7 +3,8 @@ revision = 3
 requires-python = "==3.12.*"
 
 [options]
-exclude-newer = "2026-04-13T19:47:40Z"
+exclude-newer = "2026-04-08T13:22:04.171477384Z"
```
enyst, line 6: Seems the agent noticed this is older, but it looks correct to me, that's 7 days ago indeed. WDYT?

llamantino: According to the uv docs, that's how it should be with a 7D cooldown period. I don't understand why the previous date was higher. Maybe it's related to this change in uv? https://github.com/astral-sh/uv/commit/c6cc188e03392d471ba14909eb4d7a42950cd88d

Could you please check what happens on your local repo when you use "uv lock"? My uv version is 0.11.6.

enyst: This worked and it doesn't change the file
```
uv 0.11.6 (65950801c 2026-04-09 aarch64-apple-darwin)
Resolved 400 packages in 6ms
uv lock
```

It seems we don't have a minimum/maximum version of `uv` here.

llamantino: You might need to change the pyproject.toml maybe? Try removing the prompt_toolkit line from it.

enyst: @openhands look at the commit linked from astral uv, understand the change and tell us WDYT: should the date in uv.lock be from 7 days ago, or the current day of the uv lock command? Verify and explain. Test with a newer update of some package if you can. Reply directly in-thread on github here.

enyst: I checked the Astral commit (`c6cc188`) and also reproduced this locally.

Short version: **the value stored in `uv.lock` should be the cutoff timestamp (so, for this repo, ~7 days ago), not the current day of the `uv lock` command.**

What the Astral commit changes: it fixes `uv tree --outdated` so that, when the lockfile contains a relative span, outdatedness is recomputed relative to *now*. It does **not** change how `uv lock` serializes the lockfile.

What I verified locally:
- On this PR branch, both `uv lock` with **uv 0.11.1** and `uvx --from uv==0.11.6 uv lock` are no-ops; they keep:
  - `exclude-newer = "2026-04-08...Z"`
  - `exclude-newer-span = "P7D"`
- On `origin/main`, with no dependency edits, running `uv lock` rewrites the old header from:
  - `exclude-newer = "2026-04-13T19:47:40Z"`
  to:
  - `exclude-newer = "2026-04-08...Z"`
  - `exclude-newer-span = "P7D"`
  So the previous higher date looks like older/legacy lockfile serialization, not the correct current behavior.
- In a temp repro project with `exclude-newer = "3 weeks"` + `idna`:
  - locking at simulated `2024-05-01` wrote `exclude-newer = "2024-04-10T00:00:00Z"` and resolved `idna==3.6`
  - then running `uv lock --upgrade-package idna` at simulated `2024-06-01` rewrote it to `exclude-newer = "2024-05-11T00:00:00Z"` and upgraded `idna` to `3.7`

So my take: **the PR’s `2026-04-08` value is the correct one** for `exclude-newer = "7 days"`. The linked uv change explains recomputing relative spans for commands like `tree --outdated`, but it does not imply the lockfile should store “today”.

_This comment was created by an AI assistant (OpenHands) on behalf of the user._

enyst (approved):
Thank you!

PR #675, opened 2026-04-17
Allow keyboard navigation with cursor keys in exit and switch confirmation modals

llamantino:

---

## Why

The exit confirmation dialog and the switch-conversation confirmation dialog currently support only basic Tab navigation. They do not support left/right arrow key navigation, and their focus behavior is not fully consistent. Since both dialogs represent the same two-choice confirmation flow, they should behave the same way from the keyboard.

## Summary

- Add left/right arrow key navigation to the exit and switch confirmation dialogs
- Set initial focus to the non-destructive action in both dialogs
- Keep Tab navigation and Escape dismissal behavior consistent across both dialogs

## How to Test

1. Open the exit confirmation dialog
2. Verify initial focus is on **No, dismiss**
3. Press **Left** and verify focus moves to **Yes, proceed**
4. Press **Right** and verify focus moves back to **No, dismiss**
5. Press **Tab** and verify focus still moves between the two buttons
6. Press **Escape** and verify the dialog closes without confirming exit
7. Open the switch-conversation confirmation dialog
8. Verify the same behavior there:
   - initial focus is on **No, stay**
   - **Left/Right** moves between the two actions
   - **Tab** still works
   - **Escape** dismisses the dialog without switching conversations

## Type

- [X] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


cbagwell:
After #686 was merged, there are now BINDINGS in both exit_modal and switch_conversation_model and this PR could be simplified by simply adding arrow keys to those instead of creating a new shared modal.  The Binding() and system=False will prevent it showing up at bottom of screen as a shortcut key since it's an excepted combo (same as tab not showing up which does work today).

```
        Binding("left", "focus_left_choice", "Focus left choice", system=False),
        Binding("right", "focus_right_choice", "Focus right choice", system=False),
```

PR #686, opened 2026-04-20
tui: Add ESC/Ctrl+C/Ctrl-Q key bindings to modals

### LINKED ISSUE #669: Settings screen blocks all exit keys (Ctrl+C, Ctrl+Q, Ctrl+D) — users must kill the process
## Bug Description

When the Settings screen is shown (either during initial setup for first-time users or when opened via `/settings`), **Ctrl+C, Ctrl+Q, and Ctrl+D do not work** to exit the application. Users are forced to kill the process.

## Root Cause

The `SettingsScreen` extends Textual's `ModalScreen`, which has two behaviors that block exit:

1. **Ctrl+C is intercepted** by `ModalScreen`'s built-in binding: `ctrl+c → screen.copy_text` (copies selected text — a no-op when nothing is selected). This overrides the App's `ctrl+c → request_quit` binding.
2. **Ctrl+Q and Ctrl+D don't propagate** through the modal screen — App-level bindings are not available in `ModalScreen` by default.

### Confirmed active bindings on the SettingsScreen:
```
left: action=previous_tab, node=ContentTabs
right: action=next_tab, node=ContentTabs
tab: action=focus_next, node=SettingsScreen
shift+tab: action=app.focus_previous, node=SettingsScreen
ctrl+c: action=screen.copy_text, node=SettingsScreen  ← overrides quit!
escape: action=cancel, node=SettingsScreen
```

Note: `ctrl+q` and `ctrl+d` are completely absent.

## Reproduction

1. Remove `~/.openhands/agent_settings.json` (or start with a fresh install)
2. Run `openhands`
3. The Settings screen appears
4. Press Ctrl+C → nothing happens
5. Press Ctrl+Q → nothing happens
6. Only option is to kill the process from another terminal

## Expected Behavior

Ctrl+C, Ctrl+Q, or Ctrl+D should exit the application from the Settings screen, consistent with the main app behavior.

## Proposed Fix

Add quit bindings (`ctrl+c`, `ctrl+q`, `ctrl+d`) to the `SettingsScreen` class that call `self.app.exit()`. This overrides `ModalScreen`'s default `ctrl+c → copy_text` binding and adds the missing `ctrl+q`/`ctrl+d` bindings. The same fix should also be applied to the `ExitConfirmationModal`.

_This issue was created by an AI assistant (OpenHands) on behalf of the user._

cbagwell:
Improvements to speed up keyboard interactions.

- Escape is commonly used as fast way to dismiss/abort dialog
  instead of current Tab+Enter. Add to exit/conversation modal
  to align with settings modal.
- Double Ctrl-C is also common way to acknowledge exit request
  (e.g. Toad/Gemini CLI) so add Ctrl-C to get this same behavior
  while also supporting current Enter as well. Throw in double
  Ctrl-Q for good measure.
- Add Ctrl-C/Ctrl-Q support to settings/conversation modals.
  Ctrl-C continues to mean Copy when text is selected in
  Input widgets.
- Add Footer to exit/conversation modals showing these available
  shortcuts which also aligns with settings behavior.

Fixes #669 except for Ctrl-D.  I suggest dropping support for Ctrl-D as its not commonly used and Input widgets use Ctrl-D for Delete characters to the right of cursor and can not easily be worked around (so for example, Ctrl-D will not work from prompt input box).


malhotra5 (approved):
LGTM, thank you!

PR #687, opened 2026-04-21
fix(settings): preserve native modal Tab navigation in Settings

### LINKED ISSUE #672: [UX] Settings screen does not support keyboard-only navigation - TAB key does not work
The CLI cannot be configured using only the keyboard. Users are forced to use the mouse in the settings screen because TAB does not work to move between controls when setting up the CLI for the first time.

### Problem

1. The Settings screen (shown during initial setup for first-time users) does not respond to TAB key
2. Cannot navigate between form fields using keyboard alone
3. Requires mouse interaction to click on settings options
4. Not accessible for keyboard-only users
5. Not expected for a CLI tool which should be fully keyboard-navigable

### Expected Behavior

* TAB should move focus to the next field/control
* Shift+TAB should move focus to the previous field/control
* All settings should be configurable using only the keyboard

### Additional Context

* Related to issue [APP-1313](https://linear.app/all-hands-ai/issue/APP-1313/settings-screen-blocks-all-exit-keys-ctrlc-ctrlq-ctrld-users-must-kill) (which fixed Ctrl+C/Q/D exit keys, but TAB navigation was not addressed)
* Mentioned in OpenHands CLI Roadmap discussions
* This is a UX accessibility issue for CLI users who prefer or require keyboard-only interaction

cbagwell:
- Drop the tab binding which was breaking the usage of tab within settings. Shift-tab was still working which was non-obvious to user.
- Scope the app-level tab behavior away from modal screens which resolves issue with getting stuck on input fields. The behavior doesn't make sense for any modal so not limited to settings modal.
- Mark the VirtualScroll as non-focusable to prevent needing to tab past it.

Fixes #672 - Settings screen does not support keyboard-only navigation


malhotra5 (approved):
LGTM, thank you!!

malhotra5:
@openhands please resolve the merge conflicts that the PR https://github.com/OpenHands/OpenHands-CLI/pull/686 introduced

understand both PRs carefully when resolving the merge conflicts. For the snapshot test svgs, just re-generate the snapshots instead once the merge conflict on the code changes has been resolved

PR #688, opened 2026-04-21
tui: make history panel keyboard navigable

cbagwell:
Make the history panel reliably keyboard navigable by converting to ListView.  When panel is created, it is also focused since user intent is most likely to interact with it immediately.

This updates focus behavior so keyboard users can reach and move through conversation history predictably, without hidden or misleading tab stops, and return cleanly to the prompt.

The supporting history panel updates keep the visible history state in sync with the active conversation so keyboard navigation lands on meaningful entries, including newly created conversations that are not yet persisted.

Refresh the history panel regression tests and snapshots to capture the navigation behavior using ListView and focus changes.


malhotra5 (approved):
LGTM, these are great improvements

PR #697, opened 2026-04-25
settings: clear stale max tokens on model identity change

cbagwell:
Before this change, the settings screen did not consistently treat provider, model, mode, custom model, and base_url changes as a new selected model identity.

One way to reproduce was:
- open /settings in Advanced mode with a custom base_url
- start from gemini/gemini-2.5-pro
- switch to litellm_proxy/gpt-5.3-codex
- save settings

The screen could keep a stale max_input_tokens value from the previous configuration. That stale override prevented a clean LiteLLM lookup for the newly selected model, which made the resulting save use CLI defaults rather than the new model's LiteLLM metadata.


malhotra5 (approved):
LGTM!

PR #717, opened 2026-05-06
fix: emit valid JSONL from --json mode

### LINKED ISSUE #676: bug: `--json` flag does not produce valid JSONL — emits pretty-printed blocks with `--JSON Event--` markers
## Summary

`--json` is advertised as "Streams JSONL event outputs to terminal" (see `--help`), but the implementation in `openhands_cli/utils.py` produces pretty-printed JSON blocks delimited by `--JSON Event--` markers. This is not valid JSONL and forces downstream consumers to implement a custom multi-line parser.

Closes / follow-up to [ALL-4828](https://linear.app/all-hands-ai/issue/ALL-4828/new-feature-reuqest-change-the-output-format-in-headless-mode) (which introduced `--json` but did not implement true JSONL output).

## Current behaviour

`openhands_cli/utils.py` lines 262–269:

```python
def json_callback(event: Event) -> None:
    if isinstance(event, SystemPromptEvent):
        return

    data = event.model_dump()
    pretty_json = json.dumps(data, indent=2, sort_keys=True)
    print("--JSON Event--")
    print(pretty_json)
```

Actual stdout when using `openhands --headless --json`:

```
--JSON Event--
{
  "event_type": "message",
  "sender": "agent",
  ...
}
--JSON Event--
{
  ...
}
```

There is no flag, env var, or config option to switch to single-line output — the callback is hardcoded and wired directly in `openhands_cli/tui/core/runner_factory.py:72`:

```python
event_callback: Callable[[Event], None] | None = (
    json_callback if self._json_mode else None
)
```

The existing tests in `tests/test_utils.py` assert `mock_print.call_count == 2` (header + JSON block), cementing the broken behaviour as expected.

## Expected behaviour

Each event should be emitted as a single line of valid JSON (JSONL), with no delimiter markers:

```
{"event_type": "message", "sender": "agent", ...}
{"event_type": "agent_state_changed", ...}
```

This matches the contract implied by `--help` ("Streams JSONL event outputs") and is the standard expectation for any tool, log processor, or `jq` pipeline consuming structured CLI output.

## Why this matters

The primary use case for `--json` is programmatic consumption — a parent process driving OpenHands as a subprocess and parsing its stdout stream. JSONL makes this trivial:

```python
for line in process.stdout:
    event = json.loads(line)
```

The current format requires a stateful parser to detect `--JSON Event--` markers, accumulate multi-line blocks, strip delimiters, and handle interleaved stdout noise. That is unnecessary complexity caused by the implementation not matching the documented contract.

## Proposed fix

`openhands_cli/utils.py`:

```python
def json_callback(event: Event) -> None:
    if isinstance(event, SystemPromptEvent):
        return
    print(json.dumps(event.model_dump()))
```

`tests/test_utils.py` would need updating to assert a single `print` call per event containing a valid single-line JSON string.

## References

* [ALL-4828](https://linear.app/all-hands-ai/issue/ALL-4828/new-feature-reuqest-change-the-output-format-in-headless-mode) — original feature request for clean headless output, closed when `--json` was introduced but the JSONL contract was never honoured

si-stip-tessl:
## Summary

Closes #676.

`--json` is documented as "Streams JSONL event outputs to terminal", but `json_callback` in `openhands_cli/utils.py` was emitting pretty-printed multi-line JSON blocks delimited by `--JSON Event--` header lines. This is not valid JSONL and forces downstream consumers (parent processes piping `openhands --headless --json` and parsing stdout) to implement a stateful multi-line parser instead of the standard `json.loads(line)` per stdout line.

## Change

- `openhands_cli/utils.py`: replace the header + pretty-printed block with a single `print(json.dumps(event.model_dump()))` call per event, producing one valid JSON object per line.
- `tests/test_utils.py`: update the two `json_callback` tests to assert exactly one `print` call per event and that the output contains no embedded newlines.

`SystemPromptEvent` filtering is unchanged.

## Before

```
--JSON Event--
{
  "event_type": "message",
  "sender": "agent",
  ...
}
--JSON Event--
{
  ...
}
```

## After

```
{"event_type": "message", "sender": "agent", ...}
{"event_type": "agent_state_changed", ...}
```

## Test plan

- [x] `uv run pytest tests/test_utils.py -k json_callback` — both tests pass
- [ ] Manual: run `openhands --headless --json` and confirm each line of stdout is parseable with `jq` / `json.loads`

## Notes

This re-creates #677, which was auto-closed when the original fork was deleted. Patch is identical.


xingyaoww (approved):
Thanks!

PR #732, opened 2026-05-09
fix(tui): disable markup in conversation notifications

### LINKED ISSUE #627: MarkupError when displaying error notifications containing special characters
## Bug Description

When a conversation run fails with an error that contains special characters (like `{`, `}`, `=`, `[`, `]`), the CLI crashes with a `MarkupError` instead of displaying the error notification.

This happens because error messages (e.g., Pydantic validation errors) contain raw dictionary representations that Textual's Rich markup parser interprets as invalid markup syntax.

## Error Message

```
MarkupError: Expected markup value (found '={\\'command\\': "gh api repos...: False, \\'reset\\': False}, input_type=dict]\n').
```

## Stack Trace

The error originates in Textual's toast rendering:

```
textual/widgets/_toast.py:111 in render
    Content.from_markup(notification.message)
```

Full traceback shows the path:
1. `conversation_runner.py:185` catches `ConversationRunError` and calls `_notification_callback("Conversation Error", str(e), "error")`
2. `conversation_manager.py:132` calls `self.notify(message, title=title, severity=severity)`
3. Textual's `notify()` defaults to `markup=True`, causing the message to be parsed as Rich markup
4. Special characters in the error message (from Pydantic validation errors containing dict representations) cause `MarkupError`

## Root Cause

In `openhands_cli/tui/core/conversation_manager.py` (lines 129-132):

```python
def notification_callback(
    title: str, message: str, severity: SeverityLevel
) -> None:
    self.notify(message, title=title, severity=severity)  # Missing markup=False
```

The `notify()` method defaults to `markup=True`, but error messages can contain arbitrary text with characters that Rich interprets as markup syntax.

## Proposed Fix

Pass `markup=False` to prevent Rich markup parsing of error messages:

```python
def notification_callback(
    title: str, message: str, severity: SeverityLevel
) -> None:
    self.notify(message, title=title, severity=severity, markup=False)
```

Alternatively, escape the message using `rich.markup.escape()`:

```python
from rich.markup import escape

def notification_callback(
    title: str, message: str, severity: SeverityLevel
) -> None:
    self.notify(escape(message), title=title, severity=severity)
```

## Environment

- OpenHands CLI version: 1.13.1
- Textual version: >=8.0.0,<9.0.0
- Python: 3.12

---
*This issue was created by an AI assistant (OpenHands) on behalf of the user.*

zerone0x:
## Summary
- disable Rich markup parsing for ConversationManager toast notifications
- prevent `MarkupError` when error messages contain raw braces or other special characters
- add a regression test covering the notification callback behavior

## Testing
- uv run pytest tests/tui/test_conversation_manager.py -q
- uv run ruff check openhands_cli/tui/core/conversation_manager.py tests/tui/test_conversation_manager.py

Fixes #627


enyst (approved):
Thank you!

PR #733, opened 2026-05-09
fix: allow local models without API keys

### LINKED ISSUE #673: [UX] API key still required in CLI when not needed for local LLMs
The API key is still required in the CLI, while it is no longer required in the GUI. This is confusing for local LLM users who often do not need an API key for their local LLM server.

### Problem
1. CLI prompts for API key even when using local/ollama models that do not require one
2. The GUI has been updated to not require an API key for local models
3. Inconsistent behavior between CLI and GUI causes confusion
4. Local LLM users should be able to skip entering an API key
5. Optional fields in the settings look exactly like enabled fields but are locked until an API key is entered

### Expected Behavior
- For local models (like Ollama), the API key field should be optional/hidden
- CLI should not require an API key when using a local LLM server that does not authenticate via API key
- The settings UI should clearly indicate which fields are required vs optional

### Additional Context
- Related to issue #538 which allows Vertex models in CLI without requiring LLM API key
- Partially addressed by issue #612 (AWS IAM credentials) and #538 (Vertex)
- Mentioned in OpenHands CLI Roadmap discussions
- This is a UX inconsistency between CLI and GUI behaviors

zerone0x:
## Summary
- skip API key validation for local or IAM-backed models in CLI settings
- unlock memory condensation and advanced settings for Ollama or localhost-backed models
- allow `--override-with-envs` to create agents for local/Ollama models without `LLM_API_KEY`

## Testing
- `uv run ruff check openhands_cli/stores/agent_store.py openhands_cli/tui/modals/settings/utils.py openhands_cli/tui/modals/settings/settings_screen.py tests/stores/test_env_llm_overrides.py tests/tui/modals/settings/test_settings_utils.py tests/tui/modals/settings/test_settings_screen.py`
- `uv run pytest -q tests/stores/test_env_llm_overrides.py tests/tui/modals/settings/test_settings_utils.py tests/tui/modals/settings/test_settings_screen.py`

Fixes #673

### DIFF
--- a/openhands_cli/stores/agent_store.py
+++ b/openhands_cli/stores/agent_store.py
@@ -5,6 +5,7 @@ import json
 import os
 import re
 from typing import Any
+from urllib.parse import urlparse
 
 from pydantic import BaseModel, SecretStr
 from rich.console import Console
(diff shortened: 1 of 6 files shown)


openhands_cli/stores/agent_store.py
```diff
@@ -134,6 +135,51 @@ def get_default_critic(llm: LLM, *, enable_critic: bool = True) -> CriticBase |
 ENV_LLM_MODEL = "LLM_MODEL"
 
 
+_PROVIDERS_WITH_OPTIONAL_API_KEYS = {
+    "bedrock",
+    "ollama",
+    "sagemaker",
+    "vertex_ai",
+    "vertex_ai_beta",
```
enyst, line 143: We have tried to avoid more special LLM/provider code, even though we have been a bit inconsistent. I think maybe rules for some providers, if they are really necessary, could be in the agent-sdk.

On the other hand, I thought bedrock works for no-password case. Have you tried it?

enyst (changes requested):
I think maybe this belongs in the agent-sdk if we need to do more to support this. I kinda thought it was already supported, but maybe we missed something?

Could you post logs or screenshots to show what failed and what works?

PR #738, opened 2026-05-14
[codex] Preserve explicit LLM max token overrides

neubig:
## Summary
- Treat the current settings form `max_tokens` value as an explicit `LLM.max_input_tokens` override regardless of whether the selected model changed.
- Keep empty max-token input as `None` so the SDK can use discovered/effective token limits.
- Update settings tests to cover both empty input and explicit override behavior when switching models.

## Context
This is a follow-up to OpenHands/software-agent-sdk#3255, which separates configured token-limit fields from effective discovered token limits. With that split, downstream callers should avoid inferring configured token overrides from model identity in the save layer.

## Validation
- `uv run pytest tests/tui/modals/settings/test_settings_utils.py tests/tui/modals/settings/test_settings_screen.py`
- `uv run ruff format openhands_cli/tui/modals/settings/utils.py tests/tui/modals/settings/test_settings_utils.py`
- `uv run ruff check openhands_cli/tui/modals/settings/utils.py tests/tui/modals/settings/test_settings_utils.py`
- `uv run pyright openhands_cli/tui/modals/settings/utils.py tests/tui/modals/settings/test_settings_utils.py`
- `git diff --check`

---

## 🚀 Try this PR

```bash
uvx --python 3.12 git+https://github.com/OpenHands/OpenHands-CLI.git@codex/llm-token-override-save-semantics
```

### DIFF
--- a/openhands_cli/tui/modals/settings/utils.py
+++ b/openhands_cli/tui/modals/settings/utils.py
@@ -185,14 +185,11 @@ def save_settings(
         if full_model.startswith("openhands/") and data.base_url is None:
             data.base_url = "https://llm-proxy.app.all-hands.dev/"
 
-        # Reset max_input_tokens when model changes so SDK auto-looks up
-        # from LiteLLM; preserve user-supplied value when model is unchanged.
-        max_input_tokens: int | None = None
-        if existing_agent is not None and full_model == existing_agent.llm.model:
-            if isinstance(data.max_tokens, str):
-                max_input_tokens = int(data.max_tokens)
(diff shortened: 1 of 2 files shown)


cbagwell:
I was last to touch that if statement because I've had past issues with switching from gemini to GPT models or switching providers leaving input context at 1M.  So I was interested in testing this ahead of merge and wanted to report no issues seen with this change.  Thanks!

neubig:
Great, thanks @cbagwell !

PR #740, opened 2026-05-17
feat(settings): add native Databricks Unity AI Gateway provider to settings TUI

prasadkona:
## Summary

Adds first-class support for the native **Databricks Unity AI Gateway** provider
(Databricks PWAF compliant) to the settings TUI: workspace host, auth method
selector (M2M service principal / Browser SSO U2M), live model discovery from the
workspace, and Databricks-aware env-var overrides. This gives CLI users governed
access to the foundation models served through their Databricks workspace.

Depends on the companion SDK PR (OpenHands/software-agent-sdk#3286).

---

### Settings screen additions

- **Databricks auth section** — shown only when provider is `databricks`:
  - Workspace host field
  - Auth method selector: Service Principal (M2M) / Browser SSO (U2M)
  - Context-specific credential fields (client ID + secret for M2M)
  - Dynamic auth hint generating a workspace-specific `databricks auth login`
    command for Browser SSO, using the hostname entered on the same screen

### Model discovery

- Live workspace endpoint picker (queries the Unity AI Gateway through the SDK)
  when host + credentials are available — same two-tier list as the web UI
- Graceful fallback to a static curated model list when the workspace is unreachable
- TTL cache (5 min) so switching provider tabs doesn't hammer the workspace

### Env-var overrides

- `DATABRICKS_HOST` and `DATABRICKS_TOKEN` env vars supported in headless mode
  (alongside existing `LLM_API_KEY` / `LLM_MODEL` / `LLM_BASE_URL`)

### Changes to existing files

| File | Change |
|---|---|
| `agent_store.py` | Extends `LLMEnvOverrides` for `DATABRICKS_HOST`/`DATABRICKS_TOKEN`; rebuilds Databricks instances so client state is fresh |
| `settings_screen.py` | Adds Databricks field group + auth method selector |
| `settings_tab.py` | Adds Databricks form fields (host, auth method, credentials) |
| `utils.py` | Extends `SettingsFormData` validation for Databricks auth paths |
| `choices.py` | Adds `get_databricks_model_options()` with TTL-cached live discovery |
| `model_recommendations.py` | Adds `databricks` to the provider enum |

---

## Tests

- 4 new test files: `test_choices_databricks.py`, `test_databricks_auth_method.py`,
  and additions to `test_settings_utils.py` / `test_settings_tab.py`
- Manually tested end-to-end in the TUI against a live Databricks workspace:
  M2M service principal ✅, Browser SSO (U2M) ✅

## Test plan

- [ ] `uv run pytest tests/tui/modals/settings/ -q` — all settings tests pass
- [ ] Launch CLI with `DATABRICKS_HOST=... DATABRICKS_TOKEN=... LLM_MODEL=databricks/...` — headless agent starts without prompting for API key
- [ ] Open TUI settings, select "databricks" provider, enter host → model picker populates with live endpoints

## Alignment with Databricks `ucode`

This integration follows the same credential model as
[Databricks `ucode`](https://github.com/databricks/ucode) — the *Unity AI Gateway
Coding CLI* that launches coding agents through the Databricks AI Gateway using
workspace credentials, **no API keys required**. The connector's `PROFILE` and
`UNIFIED` strategies read the workspace login a developer has already established
(`databricks auth login` / `~/.databrickscfg`), and `U2M` provides interactive
browser OAuth — so an OpenHands agent can reach AI Gateway the same key-free,
governed way `ucode` does, reusing the existing workspace session rather than a
separate token. The result is one consistent, governed path to AI Gateway (and
the Unity Catalog–governed resources behind it) across `ucode` and OpenHands.


enyst (changes requested):
Thank you for the contribution @prasadkona , I’m not sure this is the best way though, so you know: I suggest we discuss your proposal in the SDK repo PR, first.

I mean, up to you, I’m just concerned that we may or may not be able to accept your Databricks proposal, and if yes, how; and the best place for that discussion is in the SDK where we started talking.

PR #798, opened 2026-07-21
fix(utils): JSON-serialize events via `model_dump_json()` to avoid TypeError on sets

JosephMarinier:
## Why

In `--headless --json` mode, once a conversation is long enough to trigger context condensation, `json.dumps(event.model_dump())` crashes on the `Condensation` event because `forgotten_event_ids` is a Python `set`, which is not JSON serializable in `json.dumps()`.

## Change

Instead, `event.model_dump_json()` uses Pydantic's own encoder, which handles sets and other Python-native objects.

### DIFF
--- a/openhands_cli/utils.py
+++ b/openhands_cli/utils.py
@@ -1,6 +1,5 @@
 """Utility functions for LLM configuration in OpenHands CLI."""
 
-import json
 import os
 import platform
 import re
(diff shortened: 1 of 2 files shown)


JosephMarinier:
Hi @xingyaoww,
What should I do to get this PR reviewed? Thanks!

